PR #1, opened 2023-01-27
Fix arm64

chryswoods:

Fixing bugs / test failures on ARM64/aarch64

## Suggested reviewers:
@lohedges

All tests now pass on Linux/aarch64 and Mac/ARM64, except for the first test_align test (which uses fk_combu) on Mac/arm64. This is because fk_combu is giving the wrong answer on this platform. I will debug separately and fix in that repo. The fix won't change anything in BioSimSpace.


lohedges (approved):
LGTM. I had to hop on the branch and blacken things in order to get the tests to pass. It did look like the code was blackened, so I assume that you were using a different version. I'm using the version from conda-forge, currently version 22.12.0.

This makes me think that we should pinning black. I'm not sure how frequently it changes, or how often there are major changes to the formatting. (If there are, then it kind of defeats the point of autoformatting, since you'd potentially get huge diffs every now and again, where sneaky code changes could also be passed through.)

PR #2, opened 2023-02-09
Add create_workdir function to avoid code duplication.

lohedges:
# Is this pull request to introduce new functionality?

With the recent addition of file caching support I've switched to using `BioSimSpace.IO.saveMolecules` throughout, which has exposed a couple of small bugs where a mixture of absolute and relative paths have been used. In fixing those I decided to refactor the code to provide a unified way of creating a working directory. This PR adds a `BioSimSpace._Utils.create_workdir` function, which does the following:

```python
from BioSimSpace._Utils import create_workdir

work_dir, tmp_dir = create_work_dir("some/path")
```

If the user passes a `str`, then the function creates the desired working directory and returns the _absolute_ path to it, along with `None` for `tmp_dir`. If the user passes `None` to the function, then the function creates a temporary directory and returns its name, along with the associated `tempfile.TemporaryDirectory()` object.

Note that prior to adding file caching support I sometimes used the `Sire.IO` objects directly. This was because:

* Sometimes it was faster due to the extra logic in `readMolecules`.
* When the code was intitially written, the output of `readMolecules` didn't have the same file extension required by the specific engines, e.g. GROMACS, so I needed to rename. Using things like `Sire.IO.GroTop` directly allowed me to specify the extension too.
 
In addition, the way in which working directories were set up used to differ depending on the application. This is no longer the case, so it now makes sense to have a single function to avoid code duplication. 

## Checklist:

* I confirm that I have added a test for any new functionality in this pull request: [y] (All existing tests use this functionality.)
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [y] (Function docstring is exposed to Sphinx.)

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Align/_align.py
+++ b/python/BioSimSpace/Align/_align.py
@@ -39,7 +39,6 @@ import subprocess as _subprocess
 import shutil as _shutil
 import shlex as _shlex
 import sys as _sys
-import tempfile as _tempfile
 
 from .._Utils import _try_import, _have_imported, _assert_imported
 
@@ -294,20 +293,8 @@ def generateNetwork(
                 f"'n_edges_forced' must be 0 < value < {n_edges_fully_connected}."
             )
 
-    # Create a temporary working directory and store the directory name.
-    if work_dir is None:
-        tmp_dir = _tempfile.TemporaryDirectory()
-        work_dir = tmp_dir.name
-
-    # User specified working directory.
-    else:
-        # Use full path.
-        if work_dir[0] != "/":
-            work_dir = _os.getcwd() + "/" + work_dir
-
-        # Create the directory if it doesn't already exist.
-        if not _os.path.isdir(work_dir):
-            _os.makedirs(work_dir, exist_ok=True)
+    # Create the working directory.
+    work_dir, tmp_dir = _Utils.create_workdir(work_dir)
 
     # Make the LOMAP input and output directories.
     _os.makedirs(work_dir + "/inputs", exist_ok=True)
@@ -903,9 +890,8 @@ def matchAtoms(
     # Convert the timeout to seconds and take the value as an integer.
     timeout = int(timeout.seconds().value())
 
-    # Create a temporary working directory.
-    tmp_dir = _tempfile.TemporaryDirectory()
-    work_dir = tmp_dir.name
+    # Create the working directory.
+    work_dir, tmp_dir = _Utils.create_workdir()
 
     # Use RDKkit to find the maximum common substructure.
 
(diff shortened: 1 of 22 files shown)


chryswoods (approved):
Looks good. I agree it is a good idea to factor this out into this function. It fits nicely with the existing code. While not needed for now, in the future it may make things easier to combine the `tmp_dir` and `work_dir` into a single object that holds the `tmp_dir` internally (privately?) so that there is no risk of the user accidentally deleting the temporary directory before they have finished using the `work_dir`. This could be by creating a very simple class, e.g.

```
class WorkDir:
    def __init__(work_dir=None):
        if work_dir is None:
            self._tmp_dir = create temp directory
            self._work_dir = path within temp directory
        else:
            self._tmp_dir = None
            self._work_dir = path within worker

    def __str__(self):
        return self._work_dir

    def __repr__(self):
        return self._work_dir

    def is_temp_dir(self):
        return self._tmp_dir is not None
```

Now `WorkDir` will behave like a string, but it will carry the tmpdir around with it, and calling code won't have to have any knowledge that this is in a temporary directory.

lohedges:
That's a good idea. At present this is only an internal utility function anyway and the `tmp_dir` / `work_dir` that it returns are only used by internally and aren't exposed to the user. Your suggestion makes things cleaner though, and no internal objects do anything with the `tmp_dir`, other than needed to hold it for their lifetime. Will have a quick play around.

FYI: Still seeing some intermittent OpenMM import errors on Win py39 builds. Do you have access to a Windows VM to test whether we still need our fix?

chryswoods:
Yes - I will build and test on Windows this evening (I've broken my windows build so need to fix that anyway ;-))

lohedges:
Okay, I've updated by encapsulating the `work_dir` and `tmp_dir` in a class so that they aren't exposed to the user and the interface is identical either way. This seems to work nicely. Just needed to make sure that I had an `__add__` operator so that string addition worked as expected, and also update a few places in the code so that `str(work_dir)` was passed to various functions.

While the CI passes, I am still seeing repeated failures on WIndows py39. Most are due to OpenMM import errors, but occasionally I'm seeing different mappings generated using the `flexAlign` functionality. The dependency resolution was _identical_ for each CI run, so I'm not sure why the results weren't reproducible.

chryswoods (approved):
Looks great :-)

I don't know why Windows 3.9 is being so flaky. I've tried on a VM locally and don't see any problems. For now, I'd treat this as some GH Actions weirdness, and wait to see if we encounter any issues on a "real" machine.

PR #14, opened 2023-03-08
Fix molecule name when SMILES starts with "[" character

lohedges:
This PR fixes a bug caused by molecules generated from SMILES strings with a `[ ... ]` parenthesis block at the start. Since the SMILES string is used as the molecule name (for provenance) it is written out to several topology formats. This causes issues for GROMACS, where `gmx` interprets the name as a section specifier. 

As a simple fix, in cases where molecules are generated from SMILES strings, either via parameterisation, or directly using `BioSimSpace.Convert.smiles`, we check the SMILES string and add a prefix of `smiles:` if it starts with the "[" character. Ultimately we might be able to move this logic into the topology parsers themselves, but this should work for the use cases we require.

I've added tests to confirm that the molecule is renamed correctly and also taken the opportunity to replace code to create molecules from SMILES strings with our new internal functionality.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Convert/_convert.py
+++ b/python/BioSimSpace/Convert/_convert.py
@@ -87,8 +87,11 @@ def smiles(
     if not isinstance(property_map, dict):
         raise TypeError("'property_map' must be of type 'dict'.")
 
+    # Strip whitespace.
+    smiles_string = smiles_string.replace(" ", "")
+
     try:
-        return _SireWrappers.Molecule(
+        molecule = _SireWrappers.Molecule(
             _sire_smiles(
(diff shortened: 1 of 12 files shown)


chryswoods (approved):
All good :-)

lohedges:
Thanks. Annoyingly I am now seeing intermittent failures for Windows py39. This is at the `import BioSimSpace` test, with no information as to why the import failed. (Perhaps it's still the OpenMM issue?) In this case a re-run worked, but it seems to be 50/50 whether it works. Something to keep an eye on, I suppose.

PR #15, opened 2023-03-10
Increase SOMD cut-off since it uses reaction field.

lohedges:
When introducing the `BioSimSpace.Config` sub-package I took the opportunity to standardise some configuration options between the various engines, taking inspiration from the Exscientia sandpit. In doing so, I now realise that I had incorrectly lowered the non-bonded cut-off used by SOMD, which should be larger since it uses the reaction field method (by default). I've reverted this to the default value of 10 Angstrom.

## Suggested reviewers:
@chryswoods


chryswoods (approved):
Good catch - I'm trying to standardise default cutoffs now in sire - it is very painful ;-)

PR #21, opened 2023-03-13
Remove velocities when not all molecules have the property

lohedges:
This PR removes velocities from a system when molecules are combined in any way, since not all molecules in the resulting system may have velocities afterwards, or, if they do, the thermal distribution for the velocities will be invalid. This PR relies on the new `sire.legacy.IO.removeProperty` function to quickly strip the velocity property in the C++ layer.

I've also taken the opportunity to add `process.getSystem` tests for all supported engines.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/IO/_io.py
+++ b/python/BioSimSpace/IO/_io.py
@@ -672,27 +672,6 @@ def saveMolecules(filebase, system, fileformat, property_map={}, **kwargs):
     # A list of the files that have been written.
     files = []
 
-    # If not all molecules have a velocity property, then we will exclude it
-    # by re-mapping the property keyword. This avoids random data being written
-    # for missing velocities by our parsers.
-
-    # First get the user-defined name of the property.
-    vel_prop = _property_map.get("velocity", "velocity")
-    try:
-        mols_with_velocities = system.search(
-            f"mols with property {vel_prop}"
-        ).molecules()
-        num_vels = len(mols_with_velocities)
-    except:
-        num_vels = 0
-
-    # Not all atoms have velocities, remap the name so it isn't written.
-    if num_vels > 0 and num_vels != system.nMolecules():
-        _warnings.warn(
-            "Not all molecules have velocities so no velocities will be written."
-        )
-        _property_map[vel_prop] = "foo"
-
     # Save the system using each file format.
     for format in formats:
         # Copy an existing file if it exists in the cache.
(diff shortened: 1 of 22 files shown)


xiki-tempula:
I'm not quite sure of this is the best way forward. I think we do some reordering of the molecule in the amber squashing function (Which I think is to remove the molecule and add them back). @msuruzhon do you want to comment on this?

msuruzhon:
Yes, I think this is dangerous if you are combining molecules that have velocities. Maybe it makes sense to only remove the velocities if you have at least one molecule in the whole set that doesn't have them? Otherwise keep them.

lohedges:
Okay, I'll cancel this then and go back to removing velocities when not all molecules have them. I still think we need to figure out a way to deal with flagging systems that may need re-thermalising, but maybe that is something that the user will need to specify.

lohedges:
I've modified the code so velocities are only removed when not _all_ molecules have the property. I'll think more about how to flag that the system might need re-thermalising, but I imagine that this is an edge case, i.e. where the user (hopefully) knows what they're doing anyway.

xiki-tempula:
Thanks

lohedges:
No problem. My confusion was that I forget that the `squash` function is something that is done by the `Process` before running the simulation, it's not an equivalent of the `merge` function that is specific to AMBER, i.e. which case it would be done prior to running _any_ simulation, i.e. you need to repeatedly squash/unsquash for a sequence of simulations.

chryswoods (approved):
Looks good. I'm happy to accept. Reading the code reminded me that we need to talk about how much we expose the property map code to BioSimSpace. This works assuming the velocities are in the velocity property (which is fine for BioSimSpace I think?)

lohedges:
> Reading the code reminded me that we need to talk about how much we expose the property map code to BioSimSpace. 

At present I expose the property map wherever possible, since that was what I was told to do when writing the original code, i.e. you can pass a property map to most methods/functions that take a system. Internally it's only really used for the merging code, i.e. assigning properties for the lambda end states, and for finding water molecules. I do think that we might need to be careful if people start to use it more, since the ability to combine molecules means that we will need to account for situations when molecules in a system have been created/loaded using a _different_ property map. It's likely that we will need it to be a system/molecule attribute, that gets passed around and updated as things change.

chryswoods:
Yes - it is good to expose the map (and I remember saying to keep exposing it in case we ended up needing it). I did the same in sire, but it was rarely used and does create interesting edge cases. However, I am pleased it has remained exposed, as I am now finding interesting use cases for it, and it really did help with the merging code. (use cases are to supply extra parameters for functions) I raised this as a reminder that we should talk about it so that we can work out a good way to properly expose it to users in a way they can understand, and that will minimise the risk of weird edge cases... ;-)

Issue #23, opened 2023-03-13
[BUG] SireError::invalid_cast when merging a ligand with itself


msuruzhon:
Hi,

I have an issue with some [ligands](https://github.com/OpenBioSim/biosimspace/files/10958704/ligands.sdf.zip) from the OpenFF Benchmark set (so apologies I haven't reproduced it with a faster to parametrise system). The error I get when merging them is the following:

```
TypeError: SireError::invalid_cast: Cannot cast from an object of class 
"SireMol::AtomPropertyProperty" to an object of class "SireMol::AtomCharges". 
(call sire.error.get_last_error_details() for more info)
```

```python
import BioSimSpace as BSS

lig = BSS.IO.readMolecules("ligands.sdf")[0]
lig = BSS.Parameters.gaff2(lig).getMolecule()
merged_lig = BSS.Align.merge(lig, lig)
```

It seems to only happen with some ligands so it is probably something related to the input SDF file.

Many thanks.

lohedges:
It's failing trying to assign the first (and only) non-zero formal charge in the lambda = 0 state. Looking at the property keys of the ligand, the formal charge is as follows:

```python
lig._sire_object.property("formal_charge")
Out[11]:
SireMol::AtomCharges( size=55
0: 0 |e|
1: 0 |e|
2: 0 |e|
3: 0 |e|
4: 0 |e|
...
50: 0 |e|
51: 0 |e|
52: 0 |e|
53: 0 |e|
54: 0 |e|
)
```

It _is_ an `AtomCharges` property, so I'm confused by the message in the exception. Simply removing the `formal_charge` property makes things work:

```python
lig._sire_object = lig._sire_object.edit().removeProperty("formal_charge").molecule().commit()
```

I'll take a look at the SDF parser to see how `formal_charge` is being handled. It's not needed for the merge, so we should just be able to ignore it if it is proving to be problematic.

lohedges:
Okay, I think I've figured it out. Now in Sire, properties that are zero a strictly unitless when returned, e.g.:

```python
print(lig.getAtoms()[37]._sire_object.property("formal_charge"))
0
print(lig.getAtoms()[37]._sire_object.property("formal_charge").CHARGE())
0

print(lig.getAtoms()[38]._sire_object.property("formal_charge"))
1 |e|
print(lig.getAtoms()[38]._sire_object.property("formal_charge").CHARGE())
1
```

[...]

PR #24, opened 2023-03-14
Set default-valued property to avoid issues with zero values

lohedges:
This pull request fixes issue #23. It requires merging of [openbiosim/sire#25](https://github.com/OpenBioSim/sire/pull/25) before it will work. It fixes the issue caused by assigning zero-valued properties, which no-longer have any associated units. This is achieved by initialising properties in the lambda end states using default-valued properties, which are then edited atom-by-atom.

## Suggested reviewers:
@chryswoods


lohedges:
I took the opportunity to apply a few minor tweaks related to some recent issue threads. When ready, I'll merge along side #21.

PR #27, opened 2023-03-20
Remove watchdog to avoid non-deterministic parsing of AMBER output.

### LINKED ISSUE #26: [BUG] process.getRecords doesn't return anything
**Describe the bug**
So when I run a simulation with `BSS.Protocol.Production(report_interval=1)`, I think I shall get at least `runtime/timestep/report_interval` number of records in with `process.getRecords()`.

**To Reproduce**
```
import BioSimSpace.Sandpit.Exscientia as BSS
mol = BSS.Parameters.openff_unconstrained_2_0_0('C').getMolecule()
protocol = BSS.Protocol.Production(report_interval=1)
process = BSS.Process.Amber(mol.toSystem(), protocol, exe=amber_exe)
process.start()
process.wait()
process.getRecords()
```
Gives `{}`
**Expected behavior**
I shall get a dictionary in the format of 
```
{'NSTEP': ['0', '278800'],
 'TIME(PS)': ['0.000', '557.600'],
 'TEMP(K)': ['452.66', '293.49'],
 'PRESS': ['0.0', '0.0'],
 'ETOT': ['-14835.3346', '-14835.3346'],
 'EKTOT': ['5845.1152', '5845.1152'],
 'EPTOT': ['-20680.4498', '-20680.4498'],
 'BOND': ['3.1693', '3.1693'],
 'ANGLE': ['18.6614', '18.6614'],
 'DIHED': ['3.0760', '3.0760'],
 '1-4 NB': ['4.2537', '4.2537'],
 '1-4 EEL': ['-0.1536', '-0.1536'],
 'VDWAALS': ['3151.0853', '3151.0853'],
 'EELEC': ['-23860.5419', '-23860.5419'],
 'EHBOND': ['0.0000', '0.0000'],
 'RESTRAINT': ['0.0000', '0.0000'],
 'EKCMT': ['0.0000', '0.0000'],
 'VIRIAL': ['0.0000', '0.0000'],
 'VOLUME': ['64000.0000', '64000.0000'],
 'DENSITY': ['1.0122', '1.0122']}
```
Where there is `runtime/timestep/report_interval` number of records for each entry.

 - OS: Linux
 - Version of Python: 3.8
 - Version of BioSimSpace: dev
 - I confirm that I have checked this bug still exists in the latest released version of BioSimSpace: no

**Additional context**

lohedges:
This PR fixes #26 by removing the use of the [watchdog](https://github.com/gorakhargosh/watchdog) package that was used to automatically parse the AMBER energy (mdinfo) file. While this works fairly well for simulations that log data infrequently, it produces non-deterministic results when updates are rapid. For example, the `mtime` resolution on macOS is larger than on Linux, so fewer data points are recorded for a simulation reporting at the same time interval. (If the interval is small enough.) We now parse the AMBER log file manually using a similar approach to the one used for other engines that we support, i.e. using [pygtail](https://github.com/bgreenlee/pygtail) to get the updated log lines, then checking those for new records of interest. I've also added a simple test to verify that the correct number of records are recorded.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Process/_amber.py
+++ b/python/BioSimSpace/Process/_amber.py
@@ -161,9 +161,10 @@ class Amber(_process.Process):
         # Initialise the energy dictionary and header.
         self._stdout_dict = _process._MultiDict()
 
-        # Flag that there are currently no "results" generated.
+        # Initialise log file parsing flags.
         self._has_results = False
         self._finished_results = False
+        self._is_header = False
 
         # The names of the input files.
         self._rst_file = "%s/%s.rst7" % (self._work_dir, name)
@@ -1487,7 +1488,7 @@ class Amber(_process.Process):
             raise ValueError("The number of lines must be positive!")
 
         # Flag that this isn't a header line.
-        is_header = False
+        self._is_header = False
 
         # Append any new lines to the stdout list.
         for line in _pygtail.Pygtail(self._stdout_file):
@@ -1516,11 +1517,11 @@ class Amber(_process.Process):
                             # If we find a header, jump to the top of the loop.
                             if len(data) > 0:
                                 if data[0] == "NSTEP":
-                                    is_header = True
+                                    self._is_header = True
                                     continue
 
                         # Process the header record.
-                        if is_header:
+                        if self._is_header:
                             # Split the line using whitespace.
                             data = line.upper().split()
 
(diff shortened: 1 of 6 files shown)


chryswoods (dismissed):
Looks good - shame that watchdog didn't work, but better to have a unified solution on top of pygtail.

lohedges:
Yes, it's one of those things that has worked better or worse over time depending on implementation changes behind the scenes in the package. It's quite tricky, since they try to do things in a cross-platform way, so not all file system events are available on all platforms. It also seems that AMBER keeps the file open for the duration of the simulation, so it's being written as I had anticipated anyway. There are other similar solutions, but these rely on polling the file, so would also suffer from data loss depending on the polling frequency.

PR #29, opened 2023-03-27
Add ability to pass extra keyword arguments to LOMAP

### LINKED ISSUE #13: Passing kwargs to LOMAP using Align.generateNetwork()
Hi,

I'm looking to use BioSimSpace to run FEP calculations, however I would like to generate a minimal spanning graph using a radial network. This is an option in the original lomap implementation - see [example here](https://github.com/OpenFreeEnergy/Lomap/blob/main/examples/example_radial_graph.py). It appears like there's no functionality to request a radial graph using BioSimSpace, would it be straightforward to pass kwargs to LOMAP in order to achieve this?

lohedges:
This is a small update to allow users to pass extra arguments (or override defaults) through to LOMAP, which is used as the backend in `BioSImSpace.Align.generateNetwork`. The PR closes #13.

Note that this option is intended for _power users_, so we can't assume that all of the extra options that are specified will work perfectly with the logic in the rest of the function, e.g. plotting of the network (if requested), etc. I plan to see what issues come up over time and adjust the logic accordingly, e.g. saying that Y won't work if you've specified X, etc.

@chryswoods

### DIFF
--- a/python/BioSimSpace/Align/_align.py
+++ b/python/BioSimSpace/Align/_align.py
@@ -116,6 +116,7 @@ def generateNetwork(
     links_file=None,
     property_map={},
     n_edges_forced=None,
+    **kwargs,
 ):
     """
     Generate a perturbation network using Lead Optimisation Mappper (LOMAP).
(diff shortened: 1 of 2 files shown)


chryswoods (approved):
Good to have this option for power users :-)

PR #30, opened 2023-03-28
Improve protocol flexibility using mixins

lohedges:
This PR starts the work of pulling out stable features from the Exscientia sandpit and making them interoperable across the engines that we support. Here we mprove the flexibility of our simulation protocols through the use of mixins. These are internal classes, so aren't exposed directly to the user.

1) `BioSimSpace.Protocol._position_restraint_mixin._PositionRestraintMixin` allows position restraints to be used in _any_ protocol, not just equilibration.

2) `BioSimSpace.Protocol._free_energy_mixin._FreeEnergyMixin` allows non-production alchemical simulations at different lambda values, e.g. you can equilibrate at a specific lambda value. This approach was originally designed for use with AMBER, where performing minimisation/equilibration at each lambda value was found to be necessary for simulation stability. In our previous approach we allowed perturbable systems to be used with all regular protocols, where we would extract the lambda = 0/1 state prior to running. This seemed to work well enough when using SOMD or GROMACS as the FEP engine. The new mixin-based protocols, e.g. `FreeEnergyMinimisation`, are _not_ supported by SOMD, unless the lambda value is precisely 0 or 1. I've added logic to check this and convert the protocol to a non-mixin equivalent when appropriate, e.g. from `FreeEnergyMinimisation` to `Minimisation`. This is something we can address with SOMD2, which will hopefully be able to support minimisation/equilibration at specific lambda values, along with position restraints. (SOMD does have an "equlibrate to lambda", but this is not decoupled from the regular production simulation and provides no control over the runtime, etc. It is also known to be unreliable, so is normally avoided.) For consistency with old demos/tutorials, I've aliased the new `FreeEnergyProduction` protocol to `FreeEnergy`. If desired, we can deprecate the old protocol in future.

The PR also bundles a few other updates:

* Ignore `EAMBER` records when parsing AMBER output since they don't appear for all time steps and can be reconstructed from other records.
* Switch to using the Exscientia method for parsing GROMACS energy records via `gmx energy`. While this approach is less lightweight (all current records a reconstructed each call, rather than being added time-step by time-step) it allows support for a wider range of information and appears to be more robust.
* Make the Exscientia free energy analysis code compliant with `alchemlyb` versions 1 and 2. This means that we can drop `alchemlyb` as a requirement and allow the user to install whatever version they want alongside BioSimSpace.
* Switch to using native RDKit conversion throughout. When RDKit conversion was first introduced a few attributes were missing, which meant that the converted molecules couldn't be used with RDKit's FindMCS algorithm. This now works as expected and existing tests pass, meaning that we no longer need to use file based conversions in specific cases.
* Wrap the Sire-to-OpenMM conversion code. This looks more complicated than it is since I've added some extra logic to limit the types that we convert and make clear that the conversion is currently one-way. I've also clarified this in the website documentation.
* Add `py310` to our CI/CD.
* Update Sire requirement to 2023.2.0.

## Checklist

With this PR in place I should be able to begin updating the free-energy analysis code and pull across support for FEP with AMBER and support for absolute FEP simulations.

## Suggested reviewers:
@chryswoods


chryswoods (approved):
Cool - this is going to make things a lot easier. I like the integrations, e.g. with openmm etc

lohedges:
Just forgot to guard a test against GROMACS being missing. Strangely the test _did_ pass on macOS, where GROMACS wasn't installed too. Will try to figure out why this was.

PR #34, opened 2023-04-14
Update CI for new release process

lohedges:
This PR mirrors the CI changes to Sire, as required for the new release process. Since, unlike Sire, BioSimSpace is not a standalone package, i.e. it requires Sire as a base, there are a few subtle differences:

* The label of the `openbiosim` channel used during a build matches the branch against which the build is based, i.e. the `main` label for `main`, and `dev` for `devel`.
* Packages built against a particular branch are only pushed to the channel label associated with this branch, i.e. same as above, `main` for `main` and `dev` for `devel`. This is because so we can guarantee compatibility between packages with a given label. Versioning _should_ make this okay regardless, but this will ensure we don't have any headaches.

Assuming this works, I'll merge into `devel` and `main`. It's possible that I might need to fix minor issues with the CI scripts for the first builds there, but hopefully things are okay since I followed the same procedure as Sire.

I've also update the `setup.py` file for BioSimSpace so that it can handle `.dev` packages when making version comparisons.

## Checklist

## Suggested reviewers:
@chryswoods

### DIFF
--- a/.github/workflows/devel.yaml
+++ b/.github/workflows/devel.yaml
@@ -1,12 +1,10 @@
-name: Release Main
+name: Release Devel
 
 on:
   workflow_dispatch:
-    inputs:
-      upload_packages:
-        description: "Upload packages to anaconda (yes/no)?"
-        required: true
-        default: "no"
+  push:
+    branches: [ devel ]
+
 jobs:
   build:
     name: build (${{ matrix.python-version }}, ${{ matrix.platform.name }})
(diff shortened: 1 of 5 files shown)


lohedges:
I just remembered that tests might fail because of the `AmberPrm` issue. Also, I've not thought about how to version Sire correctly for the different branches, i.e. `devel` should now require `2023.3.0.dev`, whereas `main` will need `2023.2.2` when that's released.

lohedges:
Yes, the tests are segfaulting. I'll update the version Sire version requirement for this PR to `devel`, which will need to be modified when PR'ing into `main`.

chryswoods (approved):
👍

PR #35, opened 2023-04-14
Update CI for new release process on main

lohedges:
This PR brings the changes from #34 into `main`, updating the CI for the new release process. The only modification is the version of Sire against which BioSImSpace is built, i.e. currently `~=2023.2.2`. I've skipped this CI since this was tested with the `devel` PR. This should be safe to merge once the Sire 2023.2.2 release has been approved and is transferred to the `main` channel.

## Checklist

@chryswoods

### DIFF
--- a/.github/workflows/devel.yaml
+++ b/.github/workflows/devel.yaml
@@ -1,12 +1,10 @@
-name: Release Main
+name: Release Devel
 
 on:
   workflow_dispatch:
-    inputs:
-      upload_packages:
-        description: "Upload packages to anaconda (yes/no)?"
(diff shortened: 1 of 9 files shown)


lohedges:
(Note that I didn't need to back merge from `devel` but did so anyway. No features have been added to `devel`, so it was safe to do so, i.e. `main` and `devel` will be in sync with the CI fixes.)

lohedges:
I've combined the two fix branches here so that they can both be applied to `main`, i.e. they use `devel` prior to any new features.

PR #38, opened 2023-04-14
Allow readMolecules to work with a tuple of files

### LINKED ISSUE #37: [BUG] readMolecules faills when passing a tuple of files
This works:

```python
s = BSS.IO.readMolecules(["ala.top", "ala.crd"])
```

This doesn't

```python
s = BSS.IO.readMolecules(("ala.top", "ala.crd"))
```

Both used to work, but using a tuple no longer does following the switch to using the new Sire load functionality behind the scenes.

lohedges:
This pull request closes #37. Previously we could read from a list of tuple of files, now I need to convert a tuple to a list before passing through to the new Sire load functionality.

## Suggested reviewers:
 @chryswoods


chryswoods (approved):
Strange that something in sr.load doesn't like tuples, but does like lists...

PR #41, opened 2023-04-17
Fix protocol mixin inheritance

### LINKED ISSUE #40: [BUG] Protocol mixin inheritance loses base class attributes
The current order of inheritance means that the `BioSimSpace.Protocol` base class attributes are lost in derived classes. While this doesn't affect running via scripts, or any existing tests, printing a protocol within an interactive session will raise an exception due to the missing attribute.

The solution is to re-order the inheritance, so that the mixins are standalone classes and don't inherit from the base class.

lohedges:
This pull request closes #40.

## Checklist:

## Suggested reviewers:
@chryswoods


chryswoods (approved):
Thanks - I learned about mixins and better control of Python class inheritance from this change :-)

PR #51, opened 2023-04-25
Rename project test directory to "tests"

lohedges:
This PR renames the project test directory from `test` to `tests`, which is the default recommendation for `pytest`. While the old approach works fine, I've recently been playing around with some `pytest` tools, some of which make assumptions about the _default_ path for a project's test suite. Usually this can be modified, but I want to make integration as painless as possible.

## Suggested reviewers:
@chryswoods


lohedges:
Certain tests are failing since `gmx grompp` is raising an error. It looks like the tests might be running against a _newer_ version of GROMACS than previously. I'll create a new environment locally and try to debug. Everything passes with 2022.4, but this is now using 2023.1.

lohedges:
For reference, these are also tests from the Exscientia sandpit.

lohedges:
Okay, this is because the `maxwarn` option no longer allows the use of a negative number. Previously `-1` was used to specify "ignore everything", but this is no longer the case. I used to just use a large number, but switched to using `-1` since that's what I was told to use by Exscientia. Now we have the issue of not being backwards compatible, and breaking functionality mid-release cycle. Hmmm.

lohedges:
This can be fixed by reverting to using a suitably large positive number for `maxwarn`. Annoying the change in behaviour wasn't documented in the GROMACS changelog, so this is the first time I'd noticed it. This update will work with old and new versions of GROMACS.

chryswoods (dismissed):
I completely approve of this - not least because I keep forgetting it is `test` and not `tests` when I type ;-)

Issue #52, opened 2023-04-25
[BUG] process.getRecords not getting the same number of values for different columns


xiki-tempula:
**Describe the bug**
When I run `energy_dict = process.getRecords()`, I expect that for every key, there should be the same number of values.

**To Reproduce**
```
mol_1 = BSS.Parameters.openff_unconstrained_2_0_0("C").getMolecule()
mol_2 = BSS.Parameters.openff_unconstrained_2_0_0("CO").getMolecule()
merged = BSS.Align.merge(mol_1, mol_2)
solvated = BSS.Solvent.tip3p(merged, box=3 * [4 * BSS.Units.Length.nanometer])
protocol = BSS.Protocol.FreeEnergyMinimisation()
process = BSS.Process.Amber(solvated, protocol, exe=amber_gpu)
protocol.start()
process.wait()
new = process.getSystem()
protocol = BSS.Protocol.FreeEnergyEquilibration()
process = BSS.Process.Amber(new, protocol, exe=amber_gpu)
process.start()
process.wait()
energy_dict = process.getRecords()
for key in energy_dict:
    print(len(energy_dict[key]))
```
Gives 
```
1002
1002
1503
1002
1503
1503
1503
1503
1503
1503
1002
1002
1002
1002
1002
1002
1002
501
501
501
501
501
501
501
1002
501
```

**Expected behavior**
The number of values one get from each key should be the same.

lohedges:
Could you possibly uploaded the `stdout_file` used by this process, so that I don't have to re-run the simulation, which I can't do locally. Alternatively, just call `process.getOutput()` to get after the process finishes to get a zip file with the output, which you can upload. I'm assuming that the output formatting is different for a free energy simulation, so the logic used for the parsing isn't working correctly.

xiki-tempula:
[amber_output.zip](https://github.com/OpenBioSim/biosimspace/files/11322246/amber_output.zip)

xiki-tempula:
I think there will always be two TI regions for free energy claculations.

xiki-tempula:
i think you might want this parameter at the getRecord level

lohedges:
Thanks. I think I'll add some special handling for free energy protocols where I store two dictionaries, one for each TI region. If needed, I would also need to parse and store the information for the soft-core part and add methods to get back the records, since none currently exist.

lohedges:
[...]

PR #56, opened 2023-04-26
Fix amber stdout parsing

lohedges:
This PR provides a _possible_ solution for issue #52, i.e. fixing the parsing of AMBER stdout records for free-energy perturbation simulations. Previously (before #27) we parsed the `mdinfo` file using `watchdog`. This only contained a single set of records, so there wasn't any problem. In contrast, the standard output contains three sets of records, i.e. one for each degree of freedom in the system. These correspond to the two TI regions, as well as the soft core part. (In the previous case, I'm not sure if the data in the `mdinfo` file only corresponded to the first TI region, or some average of the two.)

The approach that I've take is to store a record dictionary for each degree of freedom, then allow the user to select which one they want to use by specifying the `dof` keyword argument when extracting records, which defaults to `dof=0`.

For example, to get records for the first TI region:

```python
records0 = process.getRecords(dof=0)
```

 To get the total energy or the second TI region:

```python
total_energy1 = process.getTotalEnergy(dof=1)
```

Here the degrees of freedom are indexed, with the meaning of the index specified in the docs. Regular, i.e. non-`FreeEnergy` protocols should just use the default, i.e. `dof=0`, and return nothing for other values. (This could be changed to an exception.)

In order to parse the records in a consistent way I've applied some formatting tweaks to the records keys so that the same key can be used for different degrees of freedom. This is because, due to the fixed-width nature of the output formatting, the keys can be abbreviated differently. This _could_ cause issues if you are doing some internal analysis based on the existing keys.

I've also provided some convenience functions to extract some new records, e.g. `DV/DL`. I've not exposed any of the soft-core records this way, since they aren't properly documented in the AMBER manual, so I'm not sure of their precise meaning.

I've added a unit test to check that I get the correct number of records for some example output. This test runs against `FreeEnergy` and `FreeEnergyMinimisation` protocol to ensure that the parsing works in both cases, since the formatting is different.

Things that I am still unsure of:

* In the examples that you provided the records for the two TI regions are identical in all cases, e.g.:

```
| TI region  1



   NSTEP       ENERGY          RMS            GMAX         NAME    NUMBER
      1       1.1499E+04     1.3559E+02     3.3411E+03     O        4411

 BOND    =    21543.4552  ANGLE   =        3.3520  DIHED      =        0.0000
 VDWAALS =    14714.9925  EEL     =   -24763.0744  HBOND      =        0.0000
 1-4 VDW =        0.0000  1-4 EEL =        0.0000  RESTRAINT  =        0.0000
 DV/DL  =        38.5181

| TI region  2



   NSTEP       ENERGY          RMS            GMAX         NAME    NUMBER
      1       1.1499E+04     1.3559E+02     3.3411E+03     O        4411

 BOND    =    21543.4552  ANGLE   =        3.3520  DIHED      =        0.0000
 VDWAALS =    14714.9925  EEL     =   -24763.0744  HBOND      =        0.0000
 1-4 VDW =        0.0000  1-4 EEL =        0.0000  RESTRAINT  =        0.0000
 DV/DL  =        38.5181
```

Is this just because this a contrived example, or due to it being a particular lambda value? Essentially I am wondering if you actually need the data for both regions?

* Is the output formatting the same for absolute _and_ relative free-energy simulations? If not, could you provide additional output examples.

* How would the output differ for other types of AMBER simulation that we might want to run in future? For example, would the `dof` approach always be something that works, or are we going to get into a situation where there are other types of record that would be clunky to parse in this way?

## Checklist

## Suggested reviewers:
@xiki-tempula


lohedges:
With regards to this bit...

> In order to parse the records in a consistent way I've applied some formatting tweaks to the records keys so that the same key can be used for different degrees of freedom. This is because, due to the fixed-width nature of the output formatting, the keys can be abbreviated differently. This could cause issues if you are doing some internal analysis based on the existing keys.

It should be easy for me to create a mapping between the _universal_ key and the actual record label from the output file. This would allow me to return the data with the standard AMBER format keys, which would be familiar to any AMBER user. I'll try this when I get a moment.

python/BioSimSpace/Process/_amber.py
```diff
@@ -895,7 +895,7 @@ def getElectrostaticEnergy(self, time_series=False, block="AUTO"):
         energy : :class:`Energy <BioSimSpace.Types.Energy>`
            The electrostatic energy.
         """
-        return self.getRecord("EELECT", time_series, _Units.Energy.kcal_per_mol, block)
```
xiki-tempula, line 898: Why is this change?

python/BioSimSpace/Sandpit/Exscientia/Process/_amber.py
```diff
@@ -1093,9 +1229,15 @@ def getElectrostaticEnergy(self, time_series=False, block="AUTO"):
         energy : :class:`Energy <BioSimSpace.Types.Energy>`
            The electrostatic energy.
         """
-        return self.getRecord("EELECT", time_series, _Units.Energy.kcal_per_mol, block)
+        return self.getRecord(
```
xiki-tempula, line 1232: Might add a doc to explain the different keys.

python/BioSimSpace/Process/_amber.py
lohedges: Because the key is wrong. All output that I generate for any protocol has `EELEC`, not `EELECT`. The commit message explains the reason for the change.

python/BioSimSpace/Sandpit/Exscientia/Process/_amber.py
lohedges: Good idea. Originally this was only an internal function so the user didn't really need to know what the inputs meant. Once I have the mapping in place, then I'll just explain that they are AMBER output keys and point the user to the manual. (I'm not going explain them all, since they are not even documented appropriately in the manual itself.)

python/BioSimSpace/Sandpit/Exscientia/Process/_amber.py
```diff
@@ -1410,16 +1749,31 @@ def getTotalEnergy(self, time_series=False, block="AUTO"):
         energy : :class:`Energy <BioSimSpace.Types.Energy>`
            The total energy.
         """
-        if isinstance(self._protocol, _Protocol.Minimisation):
+
+        if not isinstance(dof, int):
```
xiki-tempula, line 1753: Might get this part into a decorater instead of checking it a million times

python/BioSimSpace/Sandpit/Exscientia/Process/_amber.py
```diff
@@ -1534,9 +1932,15 @@ def getTemperature(self, time_series=False, block="AUTO"):
         temperature : :class:`Temperature <BioSimSpace.Types.Temperature>`
            The temperature.
         """
-        return self.getRecord("TEMP(K)", time_series, _Units.Temperature.kelvin, block)
+        return self.getRecord(
+            "TEMP(K)",
```
xiki-tempula, line 1936: Would the key be including unit or not including unit?

xiki-tempula (approved):
LGTM

lohedges:
Thanks. Any comments on the points raised in the original post, in particular, is the output the same for absolute and relative free-energy simulations?

xiki-tempula:
Sorry to disappoint you but I don't know the answer. To be honest, I have been using Gromacs throughout my life and has only been using amber since I joined this company.

lohedges:
No problem. I'll tag in @msuruzhon in case he knows anything more. I'm not too fussed, since you are happy with these changes, I'd just rather not add something if it will likely need to be re-worked in the near future.

Cheers.

msuruzhon:
Hi @lohedges I also don't know the answer for sure, but I think that the alchemical machinery is always the same regardless of whether RBFE or ABFE is run, so I think the output would be the same in this case. It is different for pure MD though.

lohedges:
Thanks. I'll just sort out mapping between the original and universal keys, then merge.

PR #71, opened 2023-05-19
Minor tweaks to the BioSimSpace.Stream feature

lohedges:
This PR adds some minor tweaks to the `BioSimSpace.Stream` sub-package that was re-added in #36.

* We now save using the `.bss` extension rather than `.s3` to avoid confusion with Amazon `S3` files.
* Where possible, streamed objects are now tagged with additional metadata to identify the Sire/BioSimSpace versions, along with the Sandpit from which the object was saved. This should allow the user to full recover their environment, e.g. if (for some reason) a stream doesn't load at a later date. The caveat is that we don't tag all objects that we can stream, since some are containers and don't have a property system. (We could tag all of the objects in them, or the first, but this seems like we handle the most common use cases.) I've also added a convenience function to get back any metadata associated with an object. For example:

```python
In [1]: import BioSimSpace.Sandpit.Exscientia as BSS

In [2]: mol = BSS.Parameters.openff_unconstrained_2_0_0("C").getMolecule()

In [3]: BSS.Stream.save(mol, "test")

In [4]: new_mol = BSS.Stream.load("test.bss")

In [5]: new_mol.getMetadata()
Out[5]:
{'bss_version': '2023.3.0.dev+48.gf5b4d75c',
 'sire_revisionid': 'c9331c7',
 'sandpit': 'Exscientia',
 'sire_version': '2023.3.0.dev'}
```

Note that I store both the Sire version and revision ID, since the way that Sire is versioned is different to BioSimSpace (which uses versioneer). I could try to create a similar version for Sire, but it's not trivial to work out the number of commits from the most recent tag without checkout out the repository and using `git` directly (or `pygit2`). Alternatively, it might be easier to modify the BioSimSpace version string and store a `bss_revisionid` too. That way they would both have the same formatting, along with enough information to recover the exact state. Let me know what you think.

Although we store the sandpit name, since we stream the underlying legacy Sire object it's possible to load in _any_ sandpit. I think this information is still useful, though, since properties might have been set that aren't used or exposed within a given sandpit. The same might be true if expected properties are absent. This will give a quick way to check whether you are working with an object from a compatible sandpit if something isn't working as expected. 

Let me know if there's any additional metadata that you think we should store.

## Suggested reviewers:
@chryswoods


lohedges:
Okay, it seems like an alternative approach is to take advantage of the metadata that is already written to the stream via `sire.legacy.Stream.FileHeader`, e.g.:

```python

In [2]: mol = BSS.Parameters.openff_unconstrained_2_0_0("C").getMolecule()

In [3]: import sire.legacy.Stream as Stream

In [4]: x = Stream.save(mol._sire_object)

In [5]: Stream.getDataHeader(x)
Out[5]:
*************************************
* Object type(s) : SireMol::Molecule
* Created by     : lester
* Creation date  : Fri May 19 15:32:19 2023
* Created on     : *******
* System         : Platform: Linux
*              : UNIX system: Linux
*              : UNIX release: 6.3.2-arch1-1
*              : UNIX version: #1 SMP PREEMPT_DYNAMIC Thu, 11 May 2023 16:40:42 +0000
*              : UNIX machine: x86_64
*              : Compiler: GNU C++ (12.2.0)
*              : Wordsize: 64 bit
*              : ByteOrder: Little endian
*              : Qt runtime version: 5.15.8
*              : Qt compile version: 5.15.8
*              : Sire compile version: 2023.3.0
*              : Compile flags: -fvisibility-inlines-hidden -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/lester/.conda/envs/test2/include -fdebug-prefix-map=/usr/share/miniconda3/envs/sire_build/conda-bld/sire_1683915621876/work=/usr/local/src/conda/sire-2023.3.0.dev -fdebug-prefix-map=/home/lester/.conda/envs/test2=/usr/local/src/conda-prefix  -std=c++17 -DSIRE_HAS_CPP_17 -DSIRE_HAS_CPP_14 -DSIRE_HAS_CPP_1Y -DSIRE_HAS_CPP_11 -Wall -Wno-attributes -pipe -DSIRE_ALWAYS_INLINE=inline -Wno-strict-aliasing -DSIRE_VISIBILITY_AVAILABLE -fvisibility=hidden -fvisibility-inlines-hidden -fopenmp-simd -mavx -DSIRE_USE_AVX  -std=c++17 -DSIRE_HAS_CPP_17 -DSIRE_HAS_CPP_14 -DSIRE_HAS_CPP_1Y -DSIRE_HAS_CPP_11 -O3 -ffast-math -fomit-frame-pointer
*              : Link flags: -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,--allow-shlib-undefined -Wl,-rpath,/home/lester/.conda/envs/test2/lib -Wl,-rpath-link,/home/lester/.conda/envs/test2/lib -L/home/lester/.conda/envs/test2/lib  -rdynamic -Wl,--no-undefined
* Sire Version   : 2a85e9308af8a55d7100caf906a05e100b706ef8/GitHub/2023-05-12T19:18:51+01:00 | **UNCLEAN**
* Repository     : /usr/share/miniconda3/envs/sire_build/conda-bld/git_cache/github.com/openbiosim/sire.git | devel
* Packed size    : 2.64746 kB
* Unpacked size  : 13.9297 kB
* Country        : United Kingdom
* Language       : English
*************************************
```

It could be possible to overload `Stream.save` so that additional metadata (corresponding to BioSimSpace) could be injected into the `FileHeader` for the object being streamed. This would work for any object. I'll have a think.

chryswoods:
I've added the code you need to [feature_s3_header](https://github.com/OpenBioSim/sire/pull/55). With this, add your metadata dictionary into the `FileHeader` object via 

```
sr.stream.set_header_property("bss_metadata", 
{'bss_version': '2023.3.0.dev',
 'bss_revisionid': 'gf5b4d75c',
 'sandpit': 'Exscientia',
 'sire_version': '2023.3.0.dev',
 'sire_revisionid': 'c9331c7'}
)
```

This will be saved globally and added to all s3 files written from this point on in the process.

You can get the metadata from the s3 header file the `.property()` function, e.g.

```
sr.stream.get_data_header(sr.stream.save(mol._sire_object)).property("bss_metadata")
```

lohedges:
Thanks for this. I've now wrapped the new code and added some utility functions into `BioSimSpace.Stream` to query the metadata associated with an existing stream file. These are returned as a dictionary to make querying and comparison easier. For example:

```python

INFO:numexpr.utils:Note: NumExpr detected 20 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
INFO:numexpr.utils:NumExpr defaulting to 8 threads.
WARNING:root:Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.

In [2]: mol = BSS.Parameters.openff_unconstrained_2_0_0("C").getMolecule()

In [3]: BSS.Stream.save(mol, "test.bss")

In [4]: BSS.Stream.getMetadata("test.bss")
Out[4]:
{'bss_object': 'BioSimSpace._SireWrappers._system.System',
 'bss_version': '2023.3.0.dev',
 'bss_revisionid': 'e7aa9ef',
 'sire_version': '2023.3.0.dev',
 'sire_revisionid': 'f011f58',
 'sandpit': 'None'}

In [5]: BSS.Stream.getSireMetadata("test.bss")
Out[5]:
{'Object type(s)': 'SireSystem::System',
 'Created by': 'lester',
 'Creation date': 'Mon May 22 10:25:20 2023',
 'Created on': 'porridge',
 'System': {'UNIX system': 'Linux',
  'UNIX release': '6.3.2-arch1-1',
  'UNIX version': '#1 SMP PREEMPT_DYNAMIC Thu, 11 May 2023 16:40:42 +0000',
  'UNIX machine': 'x86_64',
  'Compiler': 'GNU C++ (12.2.0)',
  'Wordsize': '64 bit',
  'ByteOrder': 'Little endian',
  'Qt runtime version': '5.15.8',
  'Qt compile version': '5.15.8',
  'Sire compile version': '2023.3.0',
  'Compile flags': '-std=c++17 -DSIRE_HAS_CPP_17 -DSIRE_HAS_CPP_14 -DSIRE_HAS_CPP_1Y -DSIRE_HAS_CPP_11 -Wall -Wno-attributes -pipe -DSIRE_ALWAYS_INLINE=inline -Wno-strict-aliasing -DSIRE_VISIBILITY_AVAILABLE -fvisibility=hidden -fvisibility-inlines-hidden -fopenmp-simd -mavx -DSIRE_USE_AVX  -std=c++17 -DSIRE_HAS_CPP_17 -DSIRE_HAS_CPP_14 -DSIRE_HAS_CPP_1Y -DSIRE_HAS_CPP_11 -O3 -ffast-math -fomit-frame-pointer',
  'Link flags': '-rdynamic -Wl,--no-undefined'},
 'Repository': 'git@github.com:OpenBioSim/sire.git | feature_s3_header',
 'Packed size': '346.805 kB',
 'Unpacked size': '3865.11 kB',
 'Country': 'United Kingdom',
 'Language': 'English'}
```

I'll re-trigger the CI when the Sire [PR](https://github.com/OpenBioSim/sire/pull/55) has been merged and the updated conda packages are available.

chryswoods:
I like the API and the conversion to dictionaries. Very clean :-)

lohedges:
I'll merge `devel` into this to re-trigger the CI now that the s3 header functionality is available in Sire.

lohedges:
That's strange, it pulled in the correct version of Sire for the tests but all of the `BioSimSpace.Stream` tests failed, but only in the Exscientia Sandpit. Strangely they all work locally. Will try to debug.

lohedges:
Ah, it's because of this:

```python
        # Extract the BioSimSpace version and revision ID.
        _bss_version = _version.get_versions()["version"].split("+")[0]
>       _bss_revisionid = _version.get_versions()["full-revisionid"][0:7]
E       TypeError: 'NoneType' object is not subscriptable
```

I wonder why this only fails in the Sandpit, and only for this PR CI run, i.e. the same thing is fine locally.

lohedges:
I'll just wrap this bit in a `try/except` block and set to `None` if it fails.

lohedges:
Okay, this is because within a conda environment the `_version.get_versions()` within the Sandpit doesn't work correctly. This works fine locally using a non-conda install, i.e. working with the development source code.

```python
In [1]: import BioSimSpace.Sandpit.Exscientia as BSS

INFO:numexpr.utils:Note: NumExpr detected 20 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
INFO:numexpr.utils:NumExpr defaulting to 8 threads.

In [2]: BSS._version.get_versions()
Out[2]:
{'version': '0+unknown',
 'full-revisionid': None,
 'dirty': None,
 'error': 'unable to compute version',
 'date': None}
```

I'll try to figure out what's gone wrong. It might just be possible to use the `_version` from the main code, but it would be nice the non user-modified files in a Sandpit were an exact mirror.

lohedges:
Yes, that works. I'll just do this:

```sh
diff --git a/python/BioSimSpace/Sandpit/Exscientia/__init__.py b/python/BioSimSpace/Sandpit/Exscientia/__init__.py
index 4a5439a8..60edd670 100644
--- a/python/BioSimSpace/Sandpit/Exscientia/__init__.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/__init__.py
@@ -244,10 +244,10 @@ from . import Trajectory
 from . import Types
 from . import Units

-from ._version import get_versions
+# Import Versioneer from the package root.
+from ... import _version

-__version__ = get_versions()["version"]
-del get_versions
+__version__ = _version.get_versions()["version"]

 import logging as _logging
```
(Along with deleting the `_version.py` file in the Sandpit.)

This actually means the `BSS.__version__` has been broken within the Sandpit from day 1 (at least when viewed from a conda download). Surprised that no-one has noticed this.

PR #89, opened 2023-05-31
Fix issue #81

### LINKED ISSUE #81: [BUG] isSame doesn't work when adding different molecules with same number of atoms to an existing system
It appears that the file cache isn't thread safe in the event of parallel threads run using `BioSimSpace.Process.ProcessRunner`. In this case, multiple threads will be calling `BioSimSpace.IO.saveMolecules`, which will access and update the cache. I believe that it is only the deletion of keys that isn't thread-safe, so it should be possible to hide this part behind a shared lock.

lohedges:
This PR closes #81.

* We now compare the `MolNum` list when comparing two systems using the `isSame` method. This catches situations where systems with the same UID contain different molecules, e.g. when they have been edited in some way, e.g. by adding a new molecule. Previously, the comparison would fail if two systems happened to contain the same number of molecules, residues, and atoms, but had _different_ molecules.
* I have also added the `MolNum` list to the file cache key, so that systems with the same UID but different molecules can be cached.
* I've added tests for both new pieces of functionality.
* While debugging I found and fixed two small bugs in the file cache. I had forgotten to decrement the total number of atoms in the cache when an entry was removed and had also forgotten to store a _copy_ of the system in the cache.

## Checklist:

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/IO/_file_cache.py
+++ b/python/BioSimSpace/IO/_file_cache.py
@@ -151,7 +151,7 @@ def check_cache(
     key = (
         system._sire_object.uid().toString(),
         format,
-        str(system._mol_nums),
+        _compress_molnum_key(str(system._mol_nums)),
         str(set(excluded_properties)),
         str(skip_water),
     )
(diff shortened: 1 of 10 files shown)


chryswoods (approved):
All good. If you wanted to compress the key further you could take a hash or similar. I don't think it would be worth doing, as I am not sure it would bring much extra performance.

lohedges:
Thanks. Yes, a hash would be a good idea. I had thought about similar but wanted the key to be somewhat human readable, mostly for debugging purposes. I guess I could use a simple, reversible hash.

PR #95, opened 2023-06-02
Fix issue #86

### LINKED ISSUE #86: [BUG] Coordinates from GROMACS vacuum simulations can cause overflow as input to AMBER
For GROMACS we implement vacuum simulations by using pseudo-PBC conditions, i.e. running calculations in a near infinite box, as described [here](https://pubmed.ncbi.nlm.nih.gov/29678588). (This approach works with all versions of GROMACS that we support.) However, a result of this is that absolute molecular coordinates can end up being really large, so can't be used as input to other engines, e.g. AMBER, since they overflow the formatting restrictions of the input file. For example:

```python

INFO:numexpr.utils:Note: NumExpr detected 20 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
INFO:numexpr.utils:NumExpr defaulting to 8 threads.
WARNING:root:Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.

In [2]: BSS.setVerbose(True)

In [3]: m = BSS.Parameters.openff_unconstrained_2_0_0("CC").getMolecule()

In [4]: p = BSS.Protocol.Minimisation()

In [5]: process = BSS.Process.Gromacs(m.toSystem(), p)
/home/lester/Code/openbiosim/biosimspace/python/BioSimSpace/Process/_gromacs.py:294: UserWarning: No simulation box found. Assuming gas phase simulation.
  _warnings.warn("No simulation box found. Assuming gas phase simulation.")
/home/lester/Code/openbiosim/biosimspace/python/BioSimSpace/_Config/_config.py:96: UserWarning: No simulation box found. Assuming gas phase simulation.
  _warnings.warn("No simulation box found. Assuming gas phase simulation.")

In [6]: process.start()
Out[6]: BioSimSpace.Process.Gromacs(<BioSimSpace.System: nMolecules=1>, BioSimSpace.Protocol.Minimisation(steps=10000, restraint=None, force_constant=10.0 kcal_per_mol/angstrom**2), exe='/home/lester/.conda/envs/openbiosim/bin.AVX2_256/gmx', name='gromacs', work_dir='/tmp/tmpyuhwqevj', seed=None)

In [7]: process.wait()

In [8]: s = process.getSystem()

In [9]: s[0].coordinates()
Out[9]:
[(9998.2400 A, 0.0200 A, 9999.0200 A),
 (9999.7600 A, -0.0200 A, 9998.9800 A),
 (9997.8600 A, -0.8300 A, 9999.6000 A),
 (9997.8900 A, 0.9500 A, 9999.4900 A),
 (9997.8300 A, -0.0400 A, 9998.0100 A),
 (1.0000e+04 A, -0.9200 A, 9998.4600 A),
 (1.0000e+04 A, 0.8600 A, 9998.4500 A),
 (1.0000e+04 A, -0.0300 A, 9999.9900 A)]

In [10]: BSS.IO.saveMolecules("test", s, "rst7")
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/lester/Code/openbiosim/biosimspace/python/BioSimSpace/IO/_io.py:730 in saveMolecules       │
│                                                                                                  │
│    727 │   │   │   │   _os.rename(file, new_file)                                                │
│    728 │   │   │   │   file = [new_file]                                                         │
│    729 │   │   │   else:                                                                         │
│ ❱  730 │   │   │   │   file = _SireIO.MoleculeParser.save(                                       │
│    731 │   │   │   │   │   system._sire_object, filebase, _property_map                          │
│    732 │   │   │   │   )                                                                         │
│    733                                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
OSError: SireError::io_error: Cannot write the (perhaps some of the ) files as the following errors occurred:
Failed to write the file '/home/lester/Downloads/vac_issue/test.rst7' using the parser for fileformat 'RST7'. Errors reported by the parser
are:
Errors converting the system to a Amber Rst7 format...
Could not write the float at index 15, value '10000.1' into the specified format AmberFormat( 6 x float[width = 12, precision = 7] ).
Could not write the float at index 18, value '10000.1' into the specified format AmberFormat( 6 x float[width = 12, precision = 7] ).
Could not write the float at index 21, value '10000.2' into the specified format AmberFormat( 6 x float[width = 12, precision = 7] ). (call
sire.error.get_last_error_details() for more info)

The above exception was the direct cause of the following exception:

╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ in <module>:1                                                                                    │
│                                                                                                  │
│ /home/lester/Code/openbiosim/biosimspace/python/BioSimSpace/IO/_io.py:742 in saveMolecules       │
│                                                                                                  │
│    739 │   │   except Exception as e:                                                            │
│    740 │   │   │   msg = "Failed to save system to format: '%s'" % format                        │
│    741 │   │   │   if _isVerbose():                                                              │
│ ❱  742 │   │   │   │   raise IOError(msg) from e                                                 │
│    743 │   │   │   else:                                                                         │
│    744 │   │   │   │   raise IOError(msg) from None                                              │
│    745                                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
OSError: Failed to save system to format: 'RST7'
```

I'm confused why this is happening since the molecule should be placed at the centre of the box, but perhaps this isn't working correctly.  For a minimisation it should be easy enough to translate it back to some reasonable place afterwards, but maybe not for other protocols where a user might want a trajectory.

lohedges:
This PR closes #86.

This is a first pass to handle molecular coordinates after vacuum simulation with GROMACS. We implement vacuum simulations by using pseudo-PBC conditions, i.e. running calculations in a near infinite box, as described [here](https://pubmed.ncbi.nlm.nih.gov/29678588), which is works on all version of GROMACS that we support. However, absolute molecular coordinates can end up being _very_ large after simulation, which causes problems when trying to write them to certain formats, i.e. the fixed-width formatting restrictions are violated.

In this PR we compute the center of mass of the _system_ following the simulation, then translate this back to the center of the box. I've needed to take care in the Exscientia Sandpit since they use two approaches read molecular frames for `getSystem`. One uses the trajectory file with wrapped coordinates, the other uses the final gro frame, in which the coordinates are unwrapped. As such, one center of mass calculation needs to use a periodic space, the other does not.

Issues that I anticipate with this approach that we might need to revisit at a later date are:

* For multi-molecule systems a center of mass translation still might end up with large absolute coordinates, e.g. if the molecules aren't close to each other. I doubt this is the typical use case for vacuum simulations, so we'll see if it becomes an issue. Here we might be able to translate molecules independently, forcing some minimum distance between them afterwards.
* This fix doesn't apply for trajectory frames obtained via `getTrajectory`. Something similar could be done here too, but the user would lose the ability to track displacements along a trajectory, etc.
* This might not work if the user sets some config option to change the wrapping during simulation. But with customised simulations a lot of our functionality is a bit hard to guarantee.

Ultimately we might want to expose a `centerMolecules` function or similar, which the user can apply before trying to write to file. This could even be an option within `saveMolecules`? Let me know if you think this would be a better solution, i.e. the user would need to opt-in, rather than us modify something. In this case I feel like they should just expect things to work, though, and it would be annoying if the behaviour was different to other engines. (I'm not sure how they deal with molecules wandering off during vacuum simulations. Perhaps they internally recenter based on the center of mass periodically?)

The center of mass calculation is currently written in Python, but can take advantage of the C++ code once it is implemented there. I've also updated the funnel metadynamics setup code to use the new functionality so as to avoid code duplication. This is tested via tutorial notebook, which I've run independently.

## Checklist

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Metadynamics/CollectiveVariable/_funnel.py
+++ b/python/BioSimSpace/Metadynamics/CollectiveVariable/_funnel.py
@@ -846,51 +846,10 @@ def makeFunnel(
     # for solvent molecules directly, since it allows for dry binding sites.
 
     # If the ligand is a Molecule, then assume the binding site is the ligand
-    # center of mass. We do this manually since Sire's built-in evaluator doesn't
-    # take into consideration molecules spanning the periodic boundary.
+    # center of mass.
     if isinstance(ligand, _Molecule):
-        # Get the "mass" property from the user map.
-        mass_prop = property_map.get("mass", "mass")
-        if mass_prop not in ligand._sire_object.propertyKeys():
-            raise _IncompatibleError("The system contains no atomic mass information!")
-
-        # Get the "coordinate" property from the user map.
-        coord_prop = property_map.get("coordinates", "coordinates")
-        if coord_prop not in ligand._sire_object.propertyKeys():
-            raise _IncompatibleError("The system contains no atomic coordinates!")
-
-        # Get the first atom in the ligand.
-        atom = ligand.getAtoms()[0]
-
-        # The sum of the atom masses.
-        total_mass = atom._sire_object.property(mass_prop).value()
-
-        # Reference coordinate.
-        ref_coord = atom._sire_object.property(coord_prop)
-
-        # Initialise the center-of-mass.
-        com = total_mass * atom._sire_object.property(coord_prop)
-
-        # Sum over the rest of the atoms.
-        for atom in ligand.getAtoms()[1:]:
-            # Get the mass and update the total.
-            mass = atom._sire_object.property(mass_prop).value()
-            total_mass += mass
-
-            # Get the coordinate and add to the reference coord using the
-            # its distance from the reference in the minimum image convention.
-            coord = atom._sire_object.property(coord_prop)
-            coord = ref_coord + _SireVector(space.calcDistVector(ref_coord, coord))
-
-            # Update the center of mass.
-            com += mass * coord
-
-        # Normalise.
-        com /= total_mass
-
-        binding_site = _Coordinate(
-            _Length(com.x(), "A"), _Length(com.y(), "A"), _Length(com.z(), "A")
-        )
+        com = ligand._getCenterOfMass(space=space, property_map=property_map)
+        binding_site = _Coordinate(*com)
 
     # Set up the grid.
 
--- a/python/BioSimSpace/Process/_gromacs.py
+++ b/python/BioSimSpace/Process/_gromacs.py
@@ -2392,12 +2392,26 @@ class Gromacs(_process.Process):
             # need to recompute it next time.
             self._mapping = mapping
 
-            # Update the box information in the original system.
-            if "space" in new_system._sire_object.propertyKeys():
+            # Get the "space" property name from the property map.
+            space_prop = self._property_map.get("space", "space")
+
+            # Update the box information in the original system. Only do this if
+            # the original system contains space information, since it will have
+            # been added in order to run vacuum simulations.
+            if (
+                space_prop in old_system._sire_object.propertyKeys()
+                and space_prop in new_system._sire_object.propertyKeys()
+            ):
                 box = new_system._sire_object.property("space")
                 old_system._sire_object.setProperty(
                     self._property_map.get("space", "space"), box
                 )
+
+            # If this is a vacuum simulation, then translate the centre of mass
+            # of the system back to the origin.
+            if not space_prop in old_system._sire_object.propertyKeys():
+                com = new_system._getCenterOfMass()
+                old_system.translate([-x for x in com])
 
             return old_system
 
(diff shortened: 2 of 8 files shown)


chryswoods (dismissed):
Thanks - I agree this should eventually be more general, as I've seen this myself in sire/openmm gas phase simulations I've run. I agree that something that periodically translates back to center is the least worst solution.

lohedges:
Good to know that this isn't isolated to GROMACS. I'll think of a more general solution if it crops up again. Bizarrely, for the test molecule here, the translation to the box edge happens within 50 steps of minimisation. It's not blown up, so I'm surprised that it's moved so far so quickly when no dynamics are being performed.

chryswoods:
That's really weird. Minimisation shouldn't give the molecule any linear momentum as it isn't sat in an external field? Certainly not enough to pull it so far outside the box. Do we see the same thing with mols.minimisation().run()? If not, then this may be worth raising with the gromacs developers as a bug.

lohedges:
I think the issue here was that the default box center for GROMACS is `(x/2, y/2, z/2)`, so I should have probably translated the system to the center before the simulation, then back again afterwards. That said, the final coordinates that I was seeing would have still implied that it had moved a long way, rather than just drifting off the edge of the periodic box. I'll test both approaches before deciding what is the best way of doing things.

PR #99, opened 2023-06-08
Fix issue #98

### LINKED ISSUE #98: [BUG] Angles will need to be updated in tests/_SireWrappers/test_system.py::test_set_box
Just noting here so that I don't forget...

When the [triclinic space fix](https://github.com/OpenBioSim/sire/pull/68) is merged into Sire we will need to update the test angles in [test_set_box](https://github.com/OpenBioSim/biosimspace/blob/50fe4c8da1ba2e121ea1eccc288578db328c2601/tests/_SireWrappers/test_system.py#L296). At the moment they are:

```python
expected_angles = [109.4712, 70.5288, 70.5288] * BSS.Units.Angle.degree
````

This needs to be changed to:

```python
expected_angles = [70.5288, 109.4712, 70.5288] * BSS.Units.Angle.degree
```

Note that the new angles are in agreement with what GROMACS expects for a default truncated octahedron, as described [here](https://manual.gromacs.org/current/reference-manual/algorithms/periodic-boundary-conditions.html). (Note that there are some typo in this page so the _exact_ angles are wrong.)

With the update to Sire we would get:

```python

In [2]: box, angles = BSS.Box.truncatedOctahedron(30*BSS.Units.Length.angstrom)

In [3]: angles
Out[3]: [70.5288 degrees, 109.4712 degrees, 70.5288 degrees]
```

lohedges:
This PR closes #98 by updating the expected box angles used in a test to those that are now generated by the triclinic lattice reduction following [this](https://github.com/OpenBioSim/sire/pull/68) fix. While I could generate the values automatically using [BioSimSpace.Box](https://biosimspace.openbiosim.org/api/index_Box.html), I'd rather have them written in so that it's easier for a user to debug. (Also, these box parameters weren't generated using functionality from that sub-package.)

## Suggested reviewers:
 @chryswoods

### DIFF
--- a/tests/Sandpit/Exscientia/_SireWrappers/test_system.py
+++ b/tests/Sandpit/Exscientia/_SireWrappers/test_system.py
@@ -292,7 +292,7 @@ def test_set_box(system):
 
     # Store the expected box dimensions and angles.
     expected_box = [30, 30, 30] * BSS.Units.Length.angstrom
-    expected_angles = [109.4712, 70.5288, 70.5288] * BSS.Units.Angle.degree
+    expected_angles = [70.5288, 109.4712, 70.5288] * BSS.Units.Angle.degree
 
     # Check that the box dimensions match.
     for b0, b1 in zip(box, expected_box):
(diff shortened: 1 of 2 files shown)


lohedges:
Looks like there's an issue with one of our dependencies on Linux py38. Will do the usual trick of trying again in the morning before investigating further.

Issue #18, opened 2023-03-10
Default SOMD Options Give Extremely Poor Performance


fjclark:
Hi,

The default SOMD `n_cycles` and `nmoves` options are resulting in very poor performance for me. Running the script shown [here](https://github.com/OpenBioSim/biosimspace/issues/17#issue-1619116837) with the GROMACS process replaced with the SOMD process results in the following simfile options:

```
save coordinates = True
ncycles = 500
nmoves = 100
ncycles_per_snap = 1
buffered coordinates frequency = 0
timestep = 2.00 femtosecond
reaction field dielectric = 78.3
cutoff type = cutoffperiodic
cutoff distance = 8 angstrom
barostat = True
pressure = 1.00000 atm
thermostat = True
temperature = 300.00 kelvin
gpu = 0
```

As a result of the extremely large number of cycles for few moves, my GPU utilisation drops from ~ 90 % (with `nmoves = 50000` and `ncycles=1`) to ~ 4 % and simulations become extremely slow. Could the number of cycles please be reduced to avoid this performance hit? Thanks.

lohedges:
Unfortunately this is the issue trying to make SOMD behave like the other engines that we support for reporting purposes, i.e. so I can have a consistent implementation of `report_interval` and `restart_interval` across all engines, so it can seem interoperable. SOMD only really performs well for _long_ simulations where you have few cycles.

This has always been an issue and I've tried to mitigate it as much as possible. We could just treat SOMD completely differently, so that it has its own defaults, but then a user might be confused why the get a different sized trajectory, or less frequent reporting. This _should_ all be fixed when we start work on SOMD2.

fjclark:
Right, thanks for the clarification. However, I still don't understand - if I set the simulation length to 5 ns, this only increases `ncycles`, while `nmoves` stays the same, meaning that performance seems equally bad:

```
save coordinates = True
ncycles = 25000
nmoves = 100
ncycles_per_snap = 1
buffered coordinates frequency = 0
timestep = 2.00 femtosecond
reaction field dielectric = 78.3
cutoff type = cutoffperiodic
cutoff distance = 8 angstrom
barostat = True
pressure = 1.00000 atm
thermostat = True
temperature = 300.00 kelvin
gpu = 0
```

lohedges:
You've just increased `runtime` without increasing `report_interval` and `restart_interval`?

jmichel80:
[...]

PR #104, opened 2023-06-15
Feature ABFE SOMD

fjclark:
The PR introduces functionality for running ABFE calculations with SOMD. This leaves the existing functionality in the Exscientia sandpit mainly untouched, other than creating an `Absolute` free energy class. This is derived from `Relative` and is almost identical other than the capability to hold restraints. 

I am very happy to add documentation once the changes to the code have been approved.

All tests pass, although see the note below.  I also tested the [ABFE notebooks](https://github.com/OpenBioSim/BioSimSpaceTutorials/tree/main/04_fep/03_ABFE) - the first runs without errors, addressing [this issue](https://github.com/OpenBioSim/BioSimSpaceTutorials/issues/11). I'm currently fixing plotting issues in the second notebook.

A side note: while I was fixing tests I noticed that [this test](https://github.com/fjclark/biosimspace-openbiosim/blob/3f1b8c1b1c923efe93a3ba1b38fd4edb4874667e/tests/Process/test_somd.py#L59) could be made to consistently fail by decreasing the ratio of cycles to moves, which produced this error:

```
Traceback (most recent call last):
  File "/home/finlayclark/anaconda3/envs/mamba/envs/bss_pr_testing/share/Sire/scripts/somd-freenrg.py", line 161, in <module>
    OpenMMMD.runFreeNrg(params)
  File "/home/finlayclark/anaconda3/envs/mamba/envs/bss_pr_testing/lib/python3.10/site-packages/sire/legacy/Tools/__init__.py", line 175, in inner
    retval = func()
  File "/home/finlayclark/anaconda3/envs/mamba/envs/bss_pr_testing/lib/python3.10/site-packages/sire/legacy/Tools/OpenMMMD.py", line 2753, in runFreeNrg
    system = moves.move(system, nmoves.val, True)
UserWarning: The periodic box size has decreased to less than twice the nonbonded cutoff.
```
I guess related to [the recent issue](https://github.com/OpenBioSim/sire/issues/49) (using Sire 2023.3.0.dev    py310h3fd9d12_58).

- Modify RestraintSearch to include a BSS-native restraint search algorithm. This depends on MDAnalysis, but as discussed we will remove this dependency in future.
- Modify Restraint to allow a more flexible analytical correction (some force constants can be 0) and add the option of a numerical correction. The numerical correction has been set to default as the analytical correction may introduce errors in some regimes.
- Allow writing of appropriate options to the somd config file/ pert file for ABFE calculations.
-  Derive an Absolute class from Relative to use for ABFE calculations. This class is extremely similar to the Relative class, but can hold restraints. This avoids running absolute calculations with Relative, which is unintuitive. However, this might be the weakest part of the code - any comments appreciated!
- Update the SOMD config options to be consistent with those in the main version of BioSimSpace (although this results in unusable default performance, [as discussed here](https://github.com/OpenBioSim/biosimspace/issues/18)).

* I confirm that I have added a test for any new functionality in this pull request: y

## Suggested reviewers:
@lohedges, @xiki-tempula


lohedges:
Thanks, @fjclark. I'm not sure why the tests failed. It's complaining about `GIT_DESCRIBE_TAG` being missing, but we have:

```
package:
  name: {{ name }}
  version: {{ GIT_DESCRIBE_TAG if GIT_DESCRIBE_TAG is not none else 'PR' }}

source:
  git_url: https://github.com/openbiosim/biosimspace.git
  git_tag: BSS_BRANCH

build:
  number: {{ GIT_DESCRIBE_NUMBER if GIT_DESCRIBE_NUMBER is not none else 0 }}
```

> A side note: while I was fixing tests I noticed that [this test](https://github.com/fjclark/biosimspace-openbiosim/blob/3f1b8c1b1c923efe93a3ba1b38fd4edb4874667e/tests/Process/test_somd.py#L59) could be made to consistently fail by decreasing the ratio of cycles to moves, which produced this error:

> Traceback (most recent call last):
  File "/home/finlayclark/anaconda3/envs/mamba/envs/bss_pr_testing/share/Sire/scripts/somd-freenrg.py", line 161, in <module>
    OpenMMMD.runFreeNrg(params)
  File "/home/finlayclark/anaconda3/envs/mamba/envs/bss_pr_testing/lib/python3.10/site-packages/sire/legacy/Tools/__init__.py", line 175, in inner
    retval = func()
  File "/home/finlayclark/anaconda3/envs/mamba/envs/bss_pr_testing/lib/python3.10/site-packages/sire/legacy/Tools/OpenMMMD.py", line 2753, in runFreeNrg
    system = moves.move(system, nmoves.val, True)
UserWarning: The periodic box size has decreased to less than twice the nonbonded cutoff.

Hmm, this is worrying... I assume the change makes it perform multiple cycles and the issue is occuring on conversions from the DCD trajectory and back. I'll see what's going on.

> Derive an Absolute class from Relative to use for ABFE calculations. This class is extremely similar to the Relative class, but can hold restraints. This avoids running absolute calculations with Relative, which is unintuitive. However, this might be the weakest part of the code - any comments appreciated!

Ultimately it would probably be best to have an `AlchemicalFreeEnergy` base class that holds common functionality with derived `Relative` and `Absolute` classes that implement the specific functionality to those methods (or overload the common methods). I'll take a look at what you've done and make some suggestions.

fjclark:
Thanks @lohedges. 

>Hmm, this is worrying... I assume the change makes it perform multiple cycles and the issue is occuring on conversions from the DCD trajectory and back. I'll see what's going on.

I'm likely misunderstanding, but this seems to be the opposite of what I found - using a single cycle with many moves consistently produced the error, while increasing the number of cycles and decreasing the number of moves per cycle consistently avoided the error (manually modifying the `.cfg` file).

lohedges:
> A side note: while I was fixing tests I noticed that [this test](https://github.com/fjclark/biosimspace-openbiosim/blob/3f1b8c1b1c923efe93a3ba1b38fd4edb4874667e/tests/Process/test_somd.py#L59) could be made to consistently fail by decreasing the ratio of cycles to moves, which produced this error:

Could you let me know what parameters you used to trigger this? I've tried a two-cycle test here and it works okay. The system is cubic too, so shouldn't be affected by the triclinic fix, unless the box information from a DCD file is always being interpreted as a general triclinic space. It's possible that you need build 154 of `2023.3.0.dev`, which isn't available for Linux yet. (There were some GCC issues.)

lohedges:
I've checked the conda docs and it looks like both `GIT_DESCRIBE_TAG` and `GIT_DESCRIBE_NUMBER` should take default values if unset, so I'm not sure what's going wrong. People have worked around this using something like `version: {{ environ.get('GIT_DESCRIBE_TAG', 'default')}}`, but this _shouldn't_ be needed. I can try adding this on `devel`, then you could merge in to your branch, and  re-push to see if it fixes things.

fjclark:
> Could you let me know what parameters you used to trigger this? 

The following works fine:

```
save coordinates = True
ncycles = 100
nmoves = 250
ncycles_per_snap = 1
buffered coordinates frequency = 0
timestep = 2.00 femtosecond
reaction field dielectric = 78.3
cutoff type = cutoffperiodic
cutoff distance = 10 angstrom
barostat = True
pressure = 1.00000 atm
thermostat = True
temperature = 300.00 kelvin
inverse friction = 1.00000 picosecond
minimise = True
constraint = hbonds-notperturbed
energy frequency = 250
lambda array = 0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0
lambda_val = 0.0
perturbed residue number = 1
```

While this produces the error:
```
save coordinates = True
ncycles = 1
nmoves = 25000
ncycles_per_snap = 1
buffered coordinates frequency = 0
timestep = 2.00 femtosecond
reaction field dielectric = 78.3
cutoff type = cutoffperiodic
cutoff distance = 10 angstrom
barostat = True
pressure = 1.00000 atm
thermostat = True
temperature = 300.00 kelvin
inverse friction = 1.00000 picosecond
minimise = True
constraint = hbonds-notperturbed
energy frequency = 250
lambda array = 0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0
lambda_val = 0.0
perturbed residue number = 1
```

lohedges:
> I'm likely misunderstanding, but this seems to be the opposite of what I found - using a single cycle with many moves consistently produced the error, while increasing the number of cycles and decreasing the number of moves per cycle consistently avoided the error (manually modifying the .cfg file).

Oh sorry, I misread. I can't think what could cause this, then, since this is just a periodic space where no reduction is being performed. Changing the number of moves/cycles shoudn't affect the box information in any way? I had (wrongly) assumed that the simulation would start running fine, then error at the second cycle when re-applying the updated box information loaded from the DCD. (This was the only place I could think that something was going wrong.)

I'll try to reproduce locally to figure out what's gone wrong. Thanks for the configuration values.

lohedges:
Just checked. It has always been NPT. I just think that this minimal test system isn't equilibrated properly.

lohedges:
> Right, thanks very much for looking into this! Why does increasing the number of cycles with constant total simulation time prevent this?

That is a good question. I'll take a look. I'm wondering whether the multi-cycle approach is using the _original_ box information when setting things up for subsequent cycles, rather than using the updated information from the DCD or OpenMM Context at the end of the previous cycle, i.e. it keeps re-setting the box so that it never blows up. If so, then this is a bug.

lohedges:
Another thought is that the barostat is never (or rarely) being deployed when running lots of short cycles, since the number of moves-per-cycle is below the attempt frequency.

lohedges:
Okay, the CI got further but still failed. Looks like we need to update our PR logic to deal with feature branches on a fork. We should be pulling a single named branch, but this doesn't  seem to be working. Will debug locally.

lohedges:
The issue is that we have this in our recipe:

```
source:
  git_url: https://github.com/openbiosim/biosimspace.git
  git_tag: BSS_BRANCH
```

so we can only build against branches of our _own_ repository. I'll figure out how to update this so that it can take the remote from the cloned fork.

lohedges:
The box vectors are being updated correctly, i.e. printing before and after the moves in each cycle are performed for the multi-cycle config:

```
###===========================================================###

###====================somd-freenrg run=======================###
Starting somd-freenrg run...
250 moves 100 cycles, 50 ps simulation time

Cycle =  1

Before: PeriodicBox( ( 30, 30, 30 ) )
After: PeriodicBox( ( 29.934, 29.934, 29.934 ) )
Backing up previous restart
Saving new restart

Cycle =  2

Before: PeriodicBox( ( 29.934, 29.934, 29.934 ) )
After: PeriodicBox( ( 30.0995, 30.0995, 30.0995 ) )
Backing up previous restart
Saving new restart

Cycle =  3

Before: PeriodicBox( ( 30.0995, 30.0995, 30.0995 ) )
After: PeriodicBox( ( 30.1397, 30.1397, 30.1397 ) )
Backing up previous restart
Saving new restart

Cycle =  4

Before: PeriodicBox( ( 30.1397, 30.1397, 30.1397 ) )
After: PeriodicBox( ( 30.1735, 30.1735, 30.1735 ) )
Backing up previous restart
Saving new restart
```

For the single cycle simulation I values like the following at the end if (it works):

```
Before: PeriodicBox( ( 30, 30, 30 ) )
After: PeriodicBox( ( 54.3718, 54.3718, 54.3718 ) )
```

Clearly the barostat working correctly in the short, multi-cycle case.

lohedges:
Looks like some tests are failing because of missing dependencies on Windows. You'll need to update the tests to guard against this. See [this](https://github.com/OpenBioSim/biosimspace/blob/devel/tests/Process/test_gromacs.py) set of unit tests for an example. It will probably involve checking for `ambertools`, `gromacs`, and `openff`, then adding something like the following around the failing tests:

```python
@pytest.mark.skipif(
    has_amber is False or has_gromacs is False or has_openff is False,
    reason="Requires AMBER, GROMACS, and OpenFF to be installed.",
)
```

python/BioSimSpace/Sandpit/Exscientia/Protocol/_free_energy_mixin.py
```diff
@@ -40,14 +40,16 @@ def __init__(
...
+           perturbation_type : str
+               The type of perturbation to perform. Options are:
+                "full" : A full perturbation of all terms (default option).
+                "discharge_soft" : Perturb all discharging soft atom charge terms (i.e. value->0.0).
+                "vanish_soft" : Perturb all vanishing soft atom LJ terms (i.e. value->0.0).
+                "flip" : Perturb all hard atom terms as well as bonds/angles.
+                "grow_soft" : Perturb all growing soft atom LJ terms (i.e. 0.0->value).
+                "charge_soft" : Perturb all charging soft atom LJ terms (i.e. 0.0->value).
+                "restraint" : Perturb the receptor-ligand restraint strength by linearly 
+                            scaling the force constants (0.0->value).
```
lohedges, line 52: Can the indentation be fixed here.

python/BioSimSpace/Sandpit/Exscientia/Protocol/_free_energy_mixin.py
```diff
@@ -96,35 +98,35 @@ def getPerturbationType(self):
...
```
lohedges, line 115: Can this be changed so that the docstring starts on its own line and the subsequent documentation is aligned with the opening parenthesis, i.e. as in the original.

python/BioSimSpace/Sandpit/Exscientia/Protocol/_free_energy_mixin.py
```diff
@@ -96,35 +98,35 @@ def getPerturbationType(self):
...
         perturbation_type = perturbation_type.lower().replace(" ", "")
 
-        allowed_perturbation_types = [
-            "full",
-            "discharge_soft",
-            "vanish_soft",
-            "flip",
-            "grow_soft",
-            "charge_soft",
-        ]
```
lohedges, line 127: This is the correct blackened formatting. Perhaps the source needs to be blackened?

python/BioSimSpace/Sandpit/Exscientia/Align/_decouple.py
```diff
@@ -1,17 +1,17 @@
...
+from .._SireWrappers import Molecule as _Molecule
 
 __all__ = ["decouple"]
 
-
-def decouple(molecule, charge=(True, False), LJ=(True, False), intramol=True):
-    """
-    Make the molecule as being decoupled, where the interactions with the
+def decouple(molecule, charge=(True, False), LJ=(True, False), property_map={}, intramol=True):
+    """Mark the molecule as being decoupled, where the interactions with the
```
lohedges, line 14: Multi-line docstrings should start on their own line.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_absolute.py
```diff
@@ -0,0 +1,166 @@
+######################################################################
+# BioSimSpace: Making biomolecular simulation a breeze!
+#
+# Copyright: 2017-2022
```
lohedges, line 4: Copyright range needs updating.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_absolute.py
```diff
@@ -0,0 +1,166 @@
...
+#
+# Authors: Lester Hedges <lester.hedges@gmail.com>
+#
+# BioSimSpace is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
```
lohedges, line 10: License version needs to be updated to `version 3`.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -311,86 +346,244 @@ def write_dihedral(key_list, equilibrium_values, force_constants):
 
         return "\n".join(output)
 
-    def toString(self, engine="Gromacs"):
-        """
-        The method for convert the restraint to a format that could be used
+    def _somd_boresch(self):
+        '''Format the SOMD string for the Boresch restraints.'''
```
lohedges, line 350: Can this be switched to double quotes, i.e. `"""`. I imagine this would be changed by `black`.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -311,86 +346,244 @@ def write_dihedral(key_list, equilibrium_values, force_constants):
...
+        restr_string += f'"equilibrium_values":{{"r0":{r0:.2f}, "thetaA0":{thetaA0:.2f}, "thetaB0":{thetaB0:.2f},"phiA0":{phiA0:.2f}, '
+        restr_string += f'"phiB0":{phiB0:.2f}, "phiC0":{phiC0:.2f}}}, '
+        restr_string += f'"force_constants":{{"kr":{kr:.2f}, "kthetaA":{kthetaA:.2f}, "kthetaB":{kthetaB:.2f}, "kphiA":{kphiA:.2f}, '
+        restr_string += f'"kphiB":{kphiB:.2f}, "kphiC":{kphiC:.2f}}}}}'
+
+        return restr_string
+
+
+    def toString(self, engine=None):
+        """The method for convert the restraint to a format that could be used
```
lohedges, line 385: Multi-line docstrings start on a new line.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -311,86 +346,244 @@ def write_dihedral(key_list, equilibrium_values, force_constants):
...
-            raise NotImplementedError(
-                f"MD Engine {engine} not implemented "
-                f"yet. Only Gromacs is supported."
-            )
+            raise NotImplementedError(f'MD Engine {engine} not implemented '
+                                      f'yet. Only Gromacs and SOMD are supported.')
+
+
+    def getCorrection(self, method="numerical"):
+        """Calculate the free energy of releasing the restraint
```
lohedges, line 413: Multi-line docstrings start on a new line.

lohedges: The indentation of the rest of the docstring will need to be adjusted accordingly.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -320,60 +336,69 @@ def workDir(self):
...
-           -------
+    def _analyse(self, restraint_type='Boresch',
+                method='MDRestraintsGenerator',
+                append_to_ligand_selection="",
+                receptor_selection_str='protein and name CA C N',
+                force_constant = None,
+                cutoff=8 * _angstrom, 
+                restraint_idx=0,
+                block='AUTO'):
+        """Analyse trajectory and select restraints which best mimic strongest
```
lohedges, line 347: Same issue with multi-line docstring.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -463,19 +492,14 @@ def _initialise_process(self, system, gpu_support, **kwargs):
...
-        """
-        Analyse existing trajectory from a simulation working directory and
+    def analyse(work_dir, system, traj, temperature, restraint_type='Boresch',
+                method='MDRestraintsGenerator',
+                append_to_ligand_selection="",
+                receptor_selection_str='protein and name CA C N',
+                force_constant=None,
+                cutoff=8 * _angstrom,
+                restraint_idx=0): 
+        """Analyse existing trajectory from a simulation working directory and
```
lohedges, line 502: Looks like this has been unblackened.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -610,29 +650,16 @@ def analyse(
...
-        method,
-        work_dir,
-        cutoff,
-    ):
-        """
-        Generate the Boresch Restraint.
+    def _boresch_restraint(u, system, temperature, lig_selection_str,
+                           recept_selection_str, method, work_dir, force_constant, 
+                           cutoff, restraint_idx=0):
+        """Generate the Boresch Restraint.
```
lohedges, line 662: Blacken.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -678,85 +722,71 @@ def _boresch_restraint(
...
-        receptor_selection_str,
-        work_dir,
-    ):
-        """
-        Generate the Boresch Restraint using MDRestraintsGenerator.
-
+    def _boresch_restraint_MDRestraintsGenerator(u, system, temperature, ligand_selection_str,
+                           receptor_selection_str, force_constant, work_dir):
+        """Generate the Boresch Restraint using MDRestraintsGenerator.
+
```
lohedges, line 745: Blacken.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -678,85 +722,71 @@ def _boresch_restraint(
...
-        Generate the Boresch Restraint using MDRestraintsGenerator.
-
+    def _boresch_restraint_MDRestraintsGenerator(u, system, temperature, ligand_selection_str,
+                           receptor_selection_str, force_constant, work_dir):
+        """Generate the Boresch Restraint using MDRestraintsGenerator.
+ 
         Parameters
         ----------
-
+
```
lohedges, line 748: Looks like some whitespace has been introduced in the following section.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -850,23 +888,28 @@ def _boresch_restraint_MDRestraintsGenerator(
...
-        ligand_selection_str,
-        receptor_selection_str,
-        work_dir,
-        cutoff,
-    ):
-        """
-        Generate the Boresch Restraint.
+    def _boresch_restraint_BSS(u, system, temperature, lig_selection_str,
+                           recept_selection_str, work_dir,  force_constant,
+                           cutoff, restraint_idx=0):
```
lohedges, line 899: Blacken.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -850,23 +888,28 @@ def _boresch_restraint_MDRestraintsGenerator(
...
+        """Generate the Boresch Restraint. This method was inspired by Irfan Alibay's
```
lohedges, line 900: Start docstring on its own line.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -906,89 +961,134 @@ def _boresch_restraint_BSS(
...
-                anchors_dict[(lig_atom_index, prot_atom_index)]["dists"].append(
-                    distance
-                )
+        
 
-        # change lists to numpy arrays
-        for pair in anchors_dict.keys():
-            anchors_dict[pair]["dists"] = _np.array(anchors_dict[pair]["dists"])
+        def _findOrderedPairs(u, lig_selection_str, recept_selection_str, cutoff):
+            """Return a list of receptor-ligand anchor atoms pairs in the form
```
lohedges, line 967: Start on new line.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -906,89 +961,134 @@ def _boresch_restraint_BSS(
...
+            pairs_ordered_sd = []
+            for item in sorted(pair_variance_dict.items(),
+                            key=lambda item: item[1]["sd"]):
+                pairs_ordered_sd.append(item[0])
+
+            return pairs_ordered_sd
+
+
+        def _getAnchorAts(a1_idx, selection_str, u):
+            """Takes in index of anchor atom 1 (in either the receptor or ligand)
```
lohedges, line 1038: Start docstring on new line.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -997,320 +1097,391 @@ def getDistance(idx1, idx2, u):
...
-                for anchor in list(
-                    anchor_idxs.keys()
-                ):  # Change to list to avoid error due to changing dict size
-                    search = mol.search(f"atomidx {anchor_idxs[anchor]}")
-                    # see if we have found the anchor
-                    if len(search) == 1:
-                        anchor_ats[anchor] = search[0]
-                        del anchor_idxs[anchor]
+        def _getConfigVol(equil_vals, force_consts, temp):
+            """Find the configurational volume accessible to the
```
lohedges, line 1135: Start docstring on new line.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -997,320 +1097,391 @@ def getDistance(idx1, idx2, u):
...
+            numerator2 = (2 * _np.pi * RT) **3 # Units: (kcal / mol )**3
+            denominator = _np.sqrt(_np.array([val for val in force_consts.values()]).prod()) 
+            # Units: (kcal / mol)**3 A**-1
+            config_vol = numerator1 * numerator2 / denominator # Units: A**3
+            return config_vol
+
+
+        def _findOrderedBoresch(u, lig_selection_str, recept_selection_str,
+                               pair_list, temp, force_constant, no_pairs=50):
+            """Calculate a list of Boresch restraints and associated
```
lohedges, line 1174: Start docstring on new line.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -997,320 +1097,391 @@ def getDistance(idx1, idx2, u):
...
+                raise _AnalysisError(
+                    "No candidate sets of Boresch restraints are suitable. Please expand "
+                    "search criteria or increase force constants.")
+            
+            return pairs_ordered_boresch, boresch_dof_data
+
+
+        def _plotDOF(ordered_restraint_labels, dof_data, restraint_idx=0,
+                    dof_to_plot=["r", "thetaA", "thetaB", "phiA", "phiB","phiC"]):
+            """Plot historgrams and variation with time of DOF over a trajectory.
```
lohedges, line 1348: Typo in `historgrams` and start docstring on new line.

python/BioSimSpace/Sandpit/Exscientia/Process/_somd.py
```diff
@@ -901,14 +939,13 @@ def _to_pert_file(
...
-        "grow_soft",
-        "charge_soft",
-    ]
+    allowed_perturbation_types = ["full",
+                                  "discharge_soft",
+                                  "vanish_soft",
+                                  "flip",
+                                  "grow_soft",
+                                  "charge_soft",
+                                  "restraint"]
```
lohedges, line 948: This has been unblackened.

lohedges (dismissed):
Other than some formatting issues (which probably just require blackening), this looks good to me. (Annoying black doesn't deal with docstring formatting issues. I've used `pydocstringformatter` before, but this messes up some specific `sphinx` tags.) I'm also not sure why our unit tests didn't flag that the code needed blackening, so will check that.

The only thing that I'm not sure about is whether _all_ attributes of `FreeEnergy.Relative` work, or are needed by, `FreeEnergy.Absolute`. If so, then deriving this way is okay, although having a shared base class, e.g. `FreeEnergy.AlchemicalFreeEnergy` would be cleaner. If not, then this could lead to confusion and broken functionality. (I assume things are okay, since the current Exs implementation just handles absolute simulations via `FreeEnergy.Relative`.)

We can think about introducing a common base class when this functionality is moved out of the sandpit.

lohedges:
Strange, `pytest --black` should have been run on Linux py39, but it doesn't seem to have happened. The logic in the conda recipe is here:

```
pytest -vvv --color=yes --black python/BioSimSpace # [linux and x86_64 and py39]
```

(This has definitely worked previously.)

lohedges:
Ah, because of this:

> The use of the Python version selectors, py27, py34, etc. is discouraged in favor of the more general comparison operators. Additional selectors in this series will not be added to conda-build.

We'll need to use `py==39`.

fjclark:
Great, thanks for going through this. I've addressed the formatting issues and blackened everything. 

@xiki-tempula, please could you let us know if you're happy for this to be merged?

lohedges:
@fjclark: Just to say that I will attempt to synchronise the Exscientia fork at some point today. I'd like to do that and merge into `devel` before merging this. It's possible that you might need to tweak your updates with respect to any changes at their end. I'll let you know when I'm done.

fjclark:
Thanks for letting me know @lohedges - I'll tweak my updates as soon as it's merged.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_absolute.py
```diff
@@ -0,0 +1,202 @@
...
+        % _analyse_freenrg
+    )
+if _sys.platform == "win32":
+    _analyse_freenrg = "%s %s" % (
+        _os.path.join(_os.path.normpath(_getBinDir()), "sire_python.exe"),
+        _analyse_freenrg,
+    )
+
+
+class Absolute(_Relative):
```
xiki-tempula, line 58: If you don't like the name of _Relative, would it be better to have a class called alchemy? It seems that `Relative` is s strict subclass of this `Absolute` class?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
```diff
@@ -1326,7 +1341,8 @@ def _initialise_runner(self, system):
                 with open(new_dir + "/somd.cfg", "r") as f:
                     for line in f:
                         if "lambda_val" in line:
-                            new_config.append("lambda_val = %s\n" % lam)
+                            # Get lam val from the Series
+                            new_config.append("lambda_val = %s\n" % lam.values[0])
```
xiki-tempula, line 1345: Might just change to the new string formatting while you are changing it.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_absolute.py
fjclark: I agree that this isn't the best way to do things. Shall I implement @lohedges's suggestion and:

1) Move all functionality to an `AlchemicalFreeEnergy` base class
2) Derive `Relative` and `Absolute` classes which are currently both identical to the base class, but allow for slightly different functionality in each in future?

This would also avoid having to update the tutorials to use the new naming and would avoid braking any current code.

fjclark: Also, do you have any comments on the above @lohedges?

xiki-tempula: My understand of class is that say if you have two `set` of functionality A and B.
The base class will be `A.intersection(B)`.
The class A will be `A - B`.
The class B will be `B - A`.

This set up makes sense if both `A - B` and `B - A` are not empty.
However, in this case, `Relative` is a strict subset of `Absolute`, so there is no functionally in `Relative` that is not used by `Absolute`.

I think it might be better to keep them in the same class and only separate when there is need to do that. In my experience, creating new classes this too early will make the later refactor very difficult.

For tutorials, I think you could just create a dummy class `Absolute(Relative)` while Relative still retain the restraint related functionality?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -326,53 +343,66 @@ def _analyse(
         method="MDRestraintsGenerator",
         append_to_ligand_selection="",
         receptor_selection_str="protein and name CA C N",
-        cutoff=10 * _angstrom,  # In Angstrom
+        force_constant=None,
+        cutoff=8 * _angstrom,
```
xiki-tempula, line 347: Why change this?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -472,7 +508,9 @@ def analyse(
         method="MDRestraintsGenerator",
         append_to_ligand_selection="",
         receptor_selection_str="protein and name CA C N",
-        cutoff=10 * _angstrom,
+        force_constant=None,
+        cutoff=8 * _angstrom,
```
xiki-tempula, line 512: Why 10?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -883,21 +1007,33 @@ def _boresch_restraint_BSS(
         temperature : :class:`System <BioSimSpace.Types.Temperature>`
             The temperature of the system
 
-        ligand_selection_str : str
+        lig_selection_str: str
             The selection string for the atoms in the ligand to consider
             as potential anchor points.
 
-        receptor_selection_str : str
+        recept_selection_str: str
```
xiki-tempula, line 1014: Might be good to keep the full description? We are not in the 8 character fortran age any more.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -906,89 +1042,145 @@ def _boresch_restraint_BSS(
...
+                    f"{recept_selection_str} and (around {cutoff / _angstrom} index {lig_atom.index})"
+                ):
+                    pair_variance_dict[(lig_atom.index, prot_atom.index)] = {}
+                    pair_variance_dict[(lig_atom.index, prot_atom.index)]["dists"] = []
+
+            # Compute Average Distance and SD
+            for frame in _tqdm(
+                u.trajectory, desc="Searching for low variance pairs. Frame no: "
+            ):
+                for lig_atom_index, prot_atom_index in pair_variance_dict.keys():
```
xiki-tempula, line 1096: Why not use https://docs.mdanalysis.org/1.1.1/documentation_pages/lib/distances.html#MDAnalysis.lib.distances.capped_distance

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -906,89 +1042,145 @@ def _boresch_restraint_BSS(
...
+                else:
+                    a3_idx = bonded_at_a2[0].index
 
             return a1_idx, a2_idx, a3_idx
 
-        ### Use These As Anchors and Plot Variance of Associated Degrees of Freedom
-
-        def getDistance(idx1, idx2, u):
+        def _getDistance(idx1, idx2, u):
             """Distance in Angstrom."""
```
xiki-tempula, line 1184: Make the doc full.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -997,71 +1189,137 @@ def getDistance(idx1, idx2, u):
             )[2][0]
             return distance
 
-        def getAngle(idx1, idx2, idx3, u):
-            """Angle in radian."""
+        def _getAngle(idx1, idx2, idx3, u):
+            """Angle in radians."""
```
xiki-tempula, line 1193: Make the doc full.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -997,71 +1189,137 @@ def getDistance(idx1, idx2, u):
...
-            BA = A - B
-            BC = C - B
-            angle = _np.arccos(_np.dot(BA, BC) / (_norm(BA) * _norm(BC)))
+            angle = _calc_angles(C, B, A, box=u.dimensions)
             return angle
 
-        def getDihedral(idx1, idx2, idx3, idx4, u):
-            """Dihedral in radian."""
+        def _getDihedral(idx1, idx2, idx3, idx4, u):
+            """Dihedral in rad"""
```
xiki-tempula, line 1201: Make the doc full.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -997,71 +1189,137 @@ def getDistance(idx1, idx2, u):
...
             positions = [u.atoms[idx].position for idx in [idx1, idx2, idx3, idx4]]
             dihedral = _calc_dihedrals(
-                positions[0],
-                positions[1],
-                positions[2],
-                positions[3],
-                box=u.dimensions,
+                positions[0], positions[1], positions[2], positions[3], box=u.dimensions
```
xiki-tempula, line 1204: I think you could just do `sum(u.atoms[idx] for idx in [idx1, idx2, idx3, idx4]).dihedral.value()`

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -997,71 +1189,137 @@ def getDistance(idx1, idx2, u):
...
-            """Angle in radian."""
+        def _getAngle(idx1, idx2, idx3, u):
+            """Angle in radians."""
             C = u.atoms[idx1].position
             B = u.atoms[idx2].position
             A = u.atoms[idx3].position
-            BA = A - B
-            BC = C - B
-            angle = _np.arccos(_np.dot(BA, BC) / (_norm(BA) * _norm(BC)))
+            angle = _calc_angles(C, B, A, box=u.dimensions)
```
xiki-tempula, line 1197: I think you could just do `sum(u.atoms[idx] for idx in [idx1, idx2, idx3]).angle.value()`

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -997,71 +1189,137 @@ def getDistance(idx1, idx2, u):
...
+            statistics over the trajectory.
+
+            Parameters
+            ----------
+
+            u : MDAnalysis.Universe
+                The trajectory for the ABFE restraint calculation as a
+                MDAnalysis.Universe object.
+
+            lig_selection_str: str
```
xiki-tempula, line 1285: Use the full name here.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -1071,215 +1329,299 @@ def getBoreschDof(l1, l2, l3, r1, r2, r3, u):
...
+                    linestyle="dashed",
+                    linewidth=2,
+                    label="mean",
+                )
+                axs[i].set_xlabel(dof_labels[dof])
+                axs[i].set_ylabel("Num Vals")
+                if i == n_dof - 1:  # Only add legend to last plot
+                    axs[i].legend()
+            fig.tight_layout()
+            fig.savefig(
```
xiki-tempula, line 1553: Destroy the figure after plotting it.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -1071,215 +1329,299 @@ def getBoreschDof(l1, l2, l3, r1, r2, r3, u):
...
+                    color="r",
+                    linestyle="dashed",
+                    linewidth=2,
+                    label="mean",
+                )  # No need to add legend as has been
+                # added to histograms
+                axs[i].set_ylabel(dof_labels[dof])
+                axs[i].set_xlabel("Frame No")
+            fig.tight_layout()
+            fig.savefig(
```
xiki-tempula, line 1576: Destroy the figure after plotting it.

python/BioSimSpace/Sandpit/Exscientia/Process/_somd.py
xiki-tempula, line 1: Didn't review this part.

python/BioSimSpace/Sandpit/Exscientia/Protocol/_config.py
```diff
@@ -447,7 +450,10 @@ def generateGromacsConfig(self, extra_options=None, extra_lines=None):
             if self.protocol.getPressure() is not None:
                 # Don't use barostat for vacuum simulations.
                 if self._has_box and self._has_water:
-                    protocol_dict["pcoupl"] = "c-rescale"  # Barostat type.
+                    if _gmx_version >= 2021:
```
xiki-tempula, line 453: Is this always safe?
Conda could give Gromacs version that is `2022.4-conda_forge`

xiki-tempula (changes requested):

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_absolute.py
fjclark: I agree that creating separate classes doesn't make sense currently, but thought it might be helpful to prevent API changes when/ if we end up needing to separate ABFE/RBFE. However, if it's better in your experience, I'm very happy to keep everything in a single class. If both you and @lohedges agree, I'll move everything back to the same class and rename it `AlchemicalFreeEnergy`.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -321,76 +428,241 @@ def toString(self, engine="Gromacs"):
...
-    def correction(self):
-        """Give the free energy of removing the restraint."""
-        if self._restraint_type == "boresch":
-            K = _k_boltz.value() * (
-                _kcal_per_mol / _kj_per_mol
-            )  # Gas constant in kJ/mol/K
-            V = (
-                (_meter3 / 1000 / _mole) / _nanometer3
-            ).value()  # standard volume in nm^3 (liter/N_A)
+    def getCorrection(self, method="numerical"):
```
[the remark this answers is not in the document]
fjclark: I changed the default to numerical to avoid any errors introduced by the analytical formula and the time penalty is negligible. Why would you prefer the analytical method? However, I've only ever seen substantial differences between the the methods when the force constants are very weak or the restraints are unstable, so happy to change the default if you'd much prefer analytical.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -151,10 +156,60 @@ def __init__(self, system, restraint_dict, temperature, restraint_type="Boresch"
...
                         f"'BioSimSpace.Types.Energy'/'BioSimSpace.Types.Angle^2'"
                     )
+
+            # Test for unstable combinations of force constants
+            non_zero_force_const = [
+                i[0]
+                for i in restraint_dict["force_constants"].items()
+                if i[1].value() != 0
+            ]
+            if "kr" not in non_zero_force_const:
```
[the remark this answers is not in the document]
fjclark: Thanks for pointing this out - I'll add tests.

My coverage looks like:

image

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
fjclark: I think 8 A seemed enough in most cases and cut the search time slightly from 10 A. However, I'll revert to 10 A for consistency.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
fjclark: Good point! Will do.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
fjclark: Thanks for pointing this out - this looks good. However, since we're aiming to replace the MDAnalysis functionality with the new sire search functionality soon anyway, I'm tempted to leave this.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
fjclark: Thanks

python/BioSimSpace/Sandpit/Exscientia/Protocol/_config.py
fjclark: This looked like it should be fine based on:
 https://github.com/OpenBioSim/biosimspace/blob/f2acfc00cb5eb9675822d6dd57f2f4c463969013/python/BioSimSpace/Sandpit/Exscientia/__init__.py#L214

xiki-tempula: Oh yeah, I added that bit.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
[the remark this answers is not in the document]
xiki-tempula: Might be good to raise the coverage a bit to >90% for all of them. Why not make this part of the CI?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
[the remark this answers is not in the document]
xiki-tempula: I think it is more like numerical integration error has hurt me too many times that I prefer to do analytical method when it is available. I think people usually go for the analytical method when they could and only switch to numerical when analytical one is not available?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
xiki-tempula: Yes, I think this would be good. Thanks.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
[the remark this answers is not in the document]
fjclark: No problem, I'll change the default to analytical. 

Just to note that I noticed this on [the Yank documentation](http://getyank.org/latest/algorithms.html#restraints-and-standard-state-correction):
>The standard state correction is computed by evaluating using a combination of numerical and analytical one-dimensional integrals from Eq. 12 of [[18]](http://getyank.org/latest/references.html#boresch2003). Note that the analytical standard state correction described in Eq. 32 of [[18]](http://getyank.org/latest/references.html#boresch2003) is inaccurate (up to several kT) in certain regimes (near raA0 and θA0, θB0 near 0 or π) and should be avoided.

>Due to numerical instabilities and the generalization of the Boresch restraints, the analytical free energy contribution is no longer supported.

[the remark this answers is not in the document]
xiki-tempula: I know that analytical correction is inaccurate when angles approaching the 0 or π. But I'm a bit surprised that numerical solution can deal with this (Without explicitly account for this)?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
fjclark: I tried replacing with the above code and this caused many of the restraint search tests to fail due to the selection of different restraints.

fjclark: Thanks, yes that was the issue (I was assuming that `calc_angles` and `calc_dihedrals` would return the same units as your code given they're all MDAnalysis functions).

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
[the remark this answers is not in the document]
fjclark: I've added the tests you mentioned as well as a few more for `_restraint_search.py`. Coverage is now:

image

`_restraint.py` is now at 90 %, and looking through the coverage report for `_restraint_search.py`, it looks like there aren't any significant gaps in the testing. I'll hang off adding tests for `_absolute.py` until we've confirmed what we're doing re the absolute and relative classes, and given that this PR targets ABFE I'd be tempted to add more tests for `_relative.py` in a different PR.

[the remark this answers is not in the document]
xiki-tempula: Thanks. Good work.

fjclark:
Thanks very much @xiki-tempula for the quick and helpful review. I think I've addressed all of your comments other than what to do about the relative and absolute classes - I'll change this tomorrow after comment from Lester.

xiki-tempula:
Thanks!

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_absolute.py
lohedges: I think that this is something that we can revisit at a later date when the functionality is moved out of the sandpit. My main concern was that you either have confusion with the naming of the class, or if there are attributes of the `Absolute` class that aren't used, or don't make sense if you aren't doing a `Relative` simulation. At present it is probably best that the implementation is consistent with what is expected in the sandpit, so I'm happy with what ever @xiki-tempula would prefer. As mentioned, we could always temporarily alias the class name for the purposes of the tutorial if it makes things clearer.

xiki-tempula (approved):
I'm happy once the `Relative` and `Absolute` class are in the same class. I don't have any preference for the name of this class.

fjclark:
> I'm happy once the `Relative` and `Absolute` class are in the same class. I don't have any preference for the name of this class.

OK, I've moved everything back into a single `AlchemicalFreeEnergy` class.

lohedges:
Is it possible to run tests locally before pushing? It would also be good to consolidate commits and only push when ready. It's a shame to waste 2.5 hours of compute time for something that could be checked locally.

Note also that this edits in this branch will need to be re-applied on top of [this](https://github.com/OpenBioSim/biosimspace/pull/122) PR once it's approved and merged into `devel`. It might be worth adding `[ci skip]` to any minor fixes until then.

Cheers.

fjclark:
> Is it possible to run tests locally before pushing? It would also be good to consolidate commits and only push when ready. It's a shame to waste 2.5 hours of compute time for something that could be checked locally.
> 
> Note also that this edits in this branch will need to be re-applied on top of [this](https://github.com/OpenBioSim/biosimspace/pull/122) PR once it's approved and merged into `devel`. It might be worth adding `[ci skip]` to any minor fixes until then.
> 
> Cheers.

Sorry, I'll make sure to test in full before pushing in future and will make sure to add `[ci skip]` to minor changes until then.

chryswoods:
A trick I use when I am continuing to develop is to create a branch from my PR, e.g. branch `feature_X2` from `feature_X`, and then I work in `feature_X2`. Pushes to `feature_X2` don't trigger the CI as they are in a different branch. When I am ready, I then either pull from `feature_X` into `feature_X2` or I close the `feature_X` PR and raise a new PR with `feature_X2`.

fjclark:
Thanks @chryswoods I'll definitely try that in future.

lohedges:
Let me know if you need help resolving the conflicts. I can always create a local copy of your branch to do this locally then PR the edits back to you.

fjclark:
Thanks a lot but it wasn't too bad. Just testing now.

lohedges:
Note that Sire packages that are needed for the CI are currently missing. I'd hold off pushing until tomorrow. No worries if not, just don't be surprised if it fails. I can re-run in the morning.

tests/Sandpit/Exscientia/FreeEnergy/test_alchemical_free_energy.py
```diff
@@ -62,7 +63,7 @@ def amber_data(tmp_path_factory):
     @pytest.fixture(scope="class")
     def gmx_complex(gmx_data):
         outdir = gmx_data
-        complex, _ = BSS.FreeEnergy.AlchemicalFreeEnergy.analyse(
+        complex, _ = BSS.FreeEnergy.Relative.analyse(
```
lohedges: Have you resolved the conflicts the correct way? It seems that `AlchemicalFreeEnergy` has been replaced by `Relative` in many places.

Also, the Sire packages are not yet available so the CI won't currently run.

fjclark: Thanks for pointing this out - I forgot to commit the final changes, sorry.

lohedges: The replacement `2023.3.0.dev` Sire packages are now available.

lohedges:
Okay, great, everything passed. I'll just check the files updated when resolving the conflicts to make sure all looks okay.

@xiki-tempula: Just to confirm, you are happy with the renaming of the `Relative` class within your sandpit? This won't cause you any issues with workflows that assume this naming, or is it easy to update? I don't really mind how things are named within the sandpit, but will revisit this at a later date when moving code out.

Cheers.

fjclark:
One last thing - I just added a test to ensure that the correct perturbed residue number is found and discovered that this is now broken. The current logic [here](https://github.com/fjclark/biosimspace-openbiosim/blob/5a15e9b13991d337fcbc08dc71079656663a4cb8/python/BioSimSpace/Sandpit/Exscientia/Protocol/_config.py#L790-L792):
```
                res_num = (
                    self.system.getDecoupledMolecules()[0]._sire_object.number().value()
                )
```
Returns `res_num` = 2 when the `LIG` is the first molecule in the `prm7` file. I think I must have broken this when I updated  the logic to be more similar to the RBFE logic at some point:
```
                res_num = (
                    self.system.search("perturbable")
                    .residues()[0]
                    ._sire_object.number()
                    .value()
```
@lohedges or @chryswoods can you suggest how I should do this correctly, please? I get a test failure if I add the line:
```
assert "perturbed residue number = 1\n" in lines
```
[here](https://github.com/fjclark/biosimspace-openbiosim/blob/5a15e9b13991d337fcbc08dc71079656663a4cb8/tests/Sandpit/Exscientia/FreeEnergy/test_alchemical_free_energy.py#L288). 

Thanks very much.

xiki-tempula:
> Just to confirm, you are happy with the renaming of the Relative class within your sandpit? This won't cause you any issues with workflows that assume this naming, or is it easy to update? I don't really mind how things are named within the sandpit, but will revisit this at a later date when moving code out.

I'm happy for this change. I could easily change the name.

lohedges:
@fjclark: This is probably because the system has been re-ordered in some way. The `ResNum` (not `ResIdx`) is preserved with the molecule. For example:

```python


In [2]: s = BSS.IO.readMolecules("amber/ala/*")

In [3]: s[0].getResidues()[0]._sire_object.number()
Out[3]: ResNum(1)

In [4]: ss = (s[1] + s[0]).toSystem()

In [5]: ss[0].getResidues()[0]._sire_object.number()
Out[5]: ResNum(4)
```

For SOMD, there is a constraint that the numbering needs to be unique and continuous, and in ascending order. I deal with this within the setup stage [here](https://github.com/OpenBioSim/biosimspace/blob/695dc480f7eca70b026d117be383f55f185ebce2/python/BioSimSpace/Process/_somd.py#L306-L307). This renumbered system is passed to the configuration generator, so that the numbering is as SOMD expects. For example, appyling this to the above:

```python


In [2]: s = BSS.IO.readMolecules("amber/ala/*")

In [3]: s[0].getResidues()[0]._sire_object.number()
Out[3]: ResNum(1)

In [4]: ss = (s[1] + s[0]).toSystem()

In [5]: ss[0].getResidues()[0]._sire_object.number()
Out[5]: ResNum(4)

In [6]: from sire.legacy.IO import renumberConstituents

In [7]: sss = BSS._SireWrappers.System(renumberConstituents(ss._sire_object))

In [8]: sss[0].getResidues()[0]._sire_object.number()
Out[8]: ResNum(1)
```

lohedges:
Note that the perturbed `ResNum` that you are seeing is probably correct. However, the system will still need to be renumbered since SOMD randomises the molecule order on startup, so there is no way to map updated coordinates and velocities back from trajectory frames unless the residue and atom numbers in the system are unique.

fjclark:
Thanks for the comments, but I'm afraid I still don't understand: if I artificially just pick the first molecule in the renumbered system by changing the code as shown below, it still has a `ResNum` of 2 rather than 1:
```
res_num = self.system.getMolecules()[0]._sire_object.number().value()
```

lohedges:
Can you point me to the test that is failing so I can try to reproduce locally. I'm not sure why the renumbering would be failing in this particular case.

fjclark:
I haven't yet pushed the test as it fails, but all I added was:

```
assert "perturbed residue number = 1\n" in lines
```

[here](https://github.com/fjclark/biosimspace-openbiosim/blob/5a15e9b13991d337fcbc08dc71079656663a4cb8/tests/Sandpit/Exscientia/FreeEnergy/test_alchemical_free_energy.py#L288).

I've also been running [this notebook](https://github.com/OpenBioSim/BioSimSpaceTutorials/blob/feature_update_abfe/04_fep/03_ABFE/01_setup_abfe.ipynb) and checking the output of Section 2.6 to see that the perturbed residue number (2) doesn't match `somd.prm7`.

lohedges:
[This](https://github.com/OpenBioSim/biosimspace/blob/695dc480f7eca70b026d117be383f55f185ebce2/tests/Process/test_somd.py#L109) test does similar, i.e. checks that the value of the perturbed residue number is changed when the molecules are re-ordered. This doesn't make sense with the renumbering, though, so I'll check what's going on.

lohedges:
I've added the following to the test:

```python
...
        print(decoupled_m.getResidues()[0]._sire_object.number())
        print(solvated[0].getResidues()[0]._sire_object.number())
        from sire.legacy.IO import renumberConstituents
        ss = BSS._SireWrappers.System(renumberConstituents(solvated._sire_object))
        print(ss[0].getResidues()[0]._sire_object.number())
...
```

This prints:

```
ResNum(1)
ResNum(1)
ResNum(1)
```

The number is definitely 1 before the system gets passed into the `AlchemicalFreeEnergy` constructor. I'll track things to figure out where it's being changed.

lohedges:
It's still `ResNum(1)` within the `Process.Somd` code before and after renumbering.

lohedges:
Ah, I see the issue:

```python
                res_num = (
                    self.system.getDecoupledMolecules()[0]._sire_object.number().value()
                )
```

This is returning a list of _molecules_ not _residues_, so you'll need to do:

```python
                res_num = (
                    self.system.getDecoupledMolecules()[0].getResidues()[0]._sire_object.number().value()
                )
```

lohedges:
Yes, that works. Here's the diff:

```
diff --git a/python/BioSimSpace/Sandpit/Exscientia/Protocol/_config.py b/python/BioSimSpace/Sandpit/Exscientia/Protocol/_config.py
index 29f0a49b..987cbb52 100644
--- a/python/BioSimSpace/Sandpit/Exscientia/Protocol/_config.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/Protocol/_config.py
@@ -788,7 +788,10 @@ class ConfigFactory:
                 )
             except ValueError:  # No perturbable molecule - try ABFE
                 res_num = (
-                    self.system.getDecoupledMolecules()[0]._sire_object.number().value()
+                    self.system.getDecoupledMolecules()[0]
+                    .getResidues()[0]
+                    ._sire_object.number()
+                    .value()
                 )
```

fjclark:
Brilliant, thanks very much - I've pushed the changes.

lohedges (approved):
Okay, great. 'll just go through the merge commit one more time to make sure that something wasn't missed. I'll merge later, or early tomorrow, then work on finalising the release.

fjclark:
Perfect, thank you.

chryswoods:
That bug would have been easier to spot if the type of the number was kept.

```
res_num = self.system.getDecoupledMolecules()[0].getResidues()[0]._sire_object.number()

assert res_num == sr.resid(1)
```

The `sr.atomid`, `sr.resid`, `sr.chainid` etc functions let you quickly create IDs of the right type. `sr.resid(1)` creates a `ResNum(1)`, while `sr.resid("ALA")` creates a `ResName("ALA")`. You can create index objects using `sr.resid(idx=1)` - this creates a `ResIdx(1)`.

lohedges:
In this case the number wont necessarily be one depending on the molecular ordering, so it's probably easiest to use `isinstance`, or do `assert res_num == sr.resid(res_num.value())`.

PR #107, opened 2023-06-16
Feature Improve Plotting

fjclark:
This PR attempts to improve the Notebook plotOverlap function:

a) To be more consistent with Fig 13 in [the best practices guide](https://livecomsjournal.org/index.php/livecoms/article/view/v2i1e18378/963). The main change is to use the same overlap cut-offs as in the best practices paper, which were not being used before. This led to misleading plots like this (despite reasonable overlap for all first off-diagonal windows) (text removed and colour bar added):

image

Where points with reasonable overlap would be grouped together with points with very low overlap. A colour bar is also added.

b) To avoid covering the whole plot in text when many windows are supplied. I've checked with 4, 16, and 36 lambda windows and the output looks reasonable in each case:

image
image
image

c) To increase flexibility so that a numpy array (as is returned from the analysis functions) can be passed directly.

This addresses comments [here](https://github.com/OpenBioSim/BioSimSpaceTutorials/issues/8).

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`):y
* I confirm that I have added a test for any new functionality in this pull request: n

## Suggested reviewers:
@lohedges

### DIFF
--- a/python/BioSimSpace/Notebook/_plot.py
+++ b/python/BioSimSpace/Notebook/_plot.py
@@ -510,7 +510,9 @@ def plotContour(x, y, z, xlabel=None, ylabel=None, zlabel=None):
     return _plt.show()
 
 
-def plotOverlapMatrix(overlap):
+def plotOverlapMatrix(overlap, 
+                      continuous_cbar=False, 
+                      color_bar_cutoffs=[0.03, 0.1, 0.3]):
     """
     Plot the overlap matrix from a free-energy perturbation analysis.
 
@@ -519,6 +521,14 @@ def plotOverlapMatrix(overlap):
 
     overlap : List of List of float, or 2D numpy array of float
         The overlap matrix.
+    continuous_cbar : bool, optional, default=False
+        If True, use a continuous colour bar. Otherwise, use a discrete
+        set of values defined by the 'color_bar_cutoffs' argument to 
+        assign a colour to each element in the matrix.
+    color_bar_cutoffs : List of float, optional, default=[0.03, 0.1, 0.3]
+        The cutoffs to use when assigning a colour to each element in the
+        matrix. This is used for both the continuous and discrete color bars.
+        Can not contain more than 3 elements.
     """
 
     # Make sure were running interactively.
@@ -538,6 +548,9 @@ def plotOverlapMatrix(overlap):
     if not isinstance(overlap, (list, tuple, _np.ndarray)):
         raise TypeError("The 'overlap' matrix must be a list of list types, or a numpy array!")
 
+    # Convert to a numpy array - no issues if this is already an array.
+    overlap = _np.array(overlap)
+
     # Store the number of rows.
     num_rows = len(overlap)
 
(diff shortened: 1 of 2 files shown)


fjclark:
Sorry, forgot to skip the CI on this and with minor tweaks to by SOMD ABFE PR - will remember in future.

lohedges:
Thanks for this. I'll hop on your branch and have a play around locally. No worries, I just skipped the CI since this isn't covered by any test.

lohedges:
Is it possible to reduce the height of the colour bar so that it matches that of the heatmap? This would avoid a bunch of whitespace to the top and bottom, which would make it much easier to include in a publication, or when making a montage with other plots.

lohedges:
I think it would look nicer if the x-axis label ran along the top, which would make the (0, 0) point the top-left corner. This appears to be what is done in the best practices paper. (The previous plot in BSS had the origin at the bottom left.)

lohedges:
It also looks like the best practices plot removes the border and sets the grid colour to white. I wonder how that would look? I guess you would still end up with the same issue if there were many windows, i.e. the white grid making things hard to see.

fjclark:
Thanks for the comments. How do these look?

image
image

fjclark:
Have also hidden the colour bar outline for consistency, e.g.


image

fjclark:
Great.

> One other thought. Do you think it could be useful to make the magic 0.3 "good overlap" cut-off a user configurable parameter with a default of 0.3? I wonder if there are situations where it would be desirable to run a quick simulation knowing there is a lower level of accuracy, or if people have different opinions as to what is an acceptable value? This is also related to @annamherz's analysis work where there is a function to check the overlap is sufficient. In that case it might also be desirable to have a user configurable option with a default of 0.3. (Having an exposed option would also make it much easier to tweak things if the "best practice" evolves over time )

Do you mean 0.03? (https://link.springer.com/article/10.1007/s10822-015-9840-9)

Personally I think I might prefer a continuous rather than a discrete colour map as blocking the 0.1 - 0.3 windows together seems quite large, although maybe the discrete colours make things clearer. Maybe we could allow the user to select discrete or continuous colour bars, and supply a list of boundaries if discrete? Or is this over-complicating things?

lohedges:
Yes 0.03 🤦‍♂️

> Personally I think I might prefer a continuous rather than a discrete colour map as blocking the 0.1 - 0.3 windows together seems quite large, although maybe the discrete colours make things clearer. Maybe we could allow the user to select discrete or continuous colour bars, and supply a list of boundaries if discrete? Or is this over-complicating things?

Yeah, not sure. As you say, perhaps it overcomplicates things, but it's probably easy enough to support. I'll have a think.

fjclark:
Ok, I've added an options for continuous or discrete color mappings, and user-specified cut-off values. The discrete default looks like:

image

While the continuous equivalent is:

image

I haven't managed to get the correspondence between the continuous and discrete bars great yet. I'm away next week so feel free to merge and modify if you have the time.

Cheers.

annamherz:
> This is also related to @annamherz's analysis work where there is a function to check the overlap is sufficient. In that case it might also be desirable to have a user configurable option with a default of 0.3. (Having an exposed option would also make it much easier to tweak things if the "best practice" evolves over time )

I think that's a good idea! I've added it for now in the [branch](https://github.com/michellab/BioSimSpace/commit/632e1962d054147af53f5dd7ad546928e868db5e).

Currently in the analysis branch, the plotting uses the plotting functions from alchemlyb for the overlap and dhdl, but I think this plotting for the overlap looks better with the colours.

PR #108, opened 2023-06-17
This fixes the Gro87/GRO87 and GroTop/GROTOP issues.

chryswoods:
This pull request includes the changes needed to allow BioSimSpace to work with the `Gro87` and `GroTop` parsers in both new sire (when they are called `GRO87` and `GROTOP`) and old sire (when they are called `Gro87` and `GroTOP`).

I've kept support for both names for now, just to ease transition. We can remove support for the older name some time in the future (e.g. `2023.4`?)

## Suggested reviewers:
@lohedges

All of the tests pass locally with the latest version of `feature_polish` (including the excluded atom fix)


lohedges:
Many thanks for this. It's ironic that I've sanitised input and used `.upper()` or `.lower()` in most places, other than for our own internal attributes. Thanks also for doing this in a backwards compatible way.

lohedges (approved):

PR #109, opened 2023-06-21
Improve unit parsing using Sire unit grammar

lohedges:
This PR improves the unit parsing within BioSimSpace by wrapping the new Sire units grammar within `BioSImSpace.Types._GeneralUnit`. When a string is passed to the constructor we first try to use the new grammar, falling back on the existing approach when this fails. I've also exposed the `.to()` method on all unit based types. This currently doesn't work correctly for temperatures, as described [here](https://github.com/OpenBioSim/sire/issues/73).

I've also fixed a few bugs that were exposed by the updates. One was an incorrect expression in a test, where parentheses were needed to ensure the order of operation was correct. The other was a parsing error in the Exs GROMACS energy record parser for surface tension records, which have different formatting to everything else. These can't be handled in the same way, so I've skipped them for now.

(Apologies for the diff in the plotting code. This is just the result of `black`. The previous code was blackened, but using the same version as the rest of the repo, i.e. `pytest-black` passed, but `black` still reformats.)

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Notebook/_plot.py
+++ b/python/BioSimSpace/Notebook/_plot.py
@@ -510,9 +510,9 @@ def plotContour(x, y, z, xlabel=None, ylabel=None, zlabel=None):
     return _plt.show()
 
 
-def plotOverlapMatrix(overlap, 
-                      continuous_cbar=False, 
-                      color_bar_cutoffs=[0.03, 0.1, 0.3]):
+def plotOverlapMatrix(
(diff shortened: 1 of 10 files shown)


chryswoods (dismissed):
Looks good. You can create a unitless unit just by passing in an integer, e.g. `GeneralUnit(1)`, `GeneralUnit("1")` or `sr.u("1")` would all be a unitless 1. Using `GeneralUnit("m/m")` works, but was confusing.

lohedges:
Ah, that makes sense. Not sure why I didn't think of trying that :man_facepalming:

PR #111, opened 2023-06-21
Fix issue #84

### LINKED ISSUE #84: [BUG] AMBER Process getPressure would not recognise
**Describe the bug**
The first frame of my amber output says `PRESS =********`, which is ignored in Process.getPressure, this would mean that getPressure would return a different number of elements compared with other metrics.

**To Reproduce**

```
import BioSimSpace.Sandpit.Exscientia as BSS
mol = BSS.Parameters.openff_unconstrained_2_0_0('C').getMolecule()
process = BSS.Process.Amber(mol.toSystem(), BSS.Protocol.FreeEnergyEquilibration(), work_dir='.')
print(len(self.getPressure(True)))
print(len(self.self.getVolume(True)))
```

There is 26 numbers in Volume but only 25 in pressure.

**Expected behavior**
Both of them give 26 numbers

**Input files**
[amber.out.zip](https://github.com/OpenBioSim/biosimspace/files/11567204/amber.out.zip)

lohedges:
This PR closes #84. We now handle missing values in AMBER output files by extending the regex to parse records such as `PRESS =********`. In all cases, any missing values are interpreted as `None`. I've modified one of the files used in the AMBER parsing tests to contain a missing value in order to confirm that things work as expected. Currently this code is only applied to the Exscientia Sandpit, where the records are being used. (They are saved to disk after a run.) I'll move this into the core code when I port over the AMBER FEP code.

I've also updated the plotting code to handle `None` values in time-series data with units. When `None` is present we filter the both the x and y data (and any associated errors) to remove data corresponding to the `None` indices in _both_ sets, i.e. if `None` is present in x, then we remove this value from x and the same index from y. I've not updated the logic for raw `float` values since this isn't used internally by BioSimSpace, nor the code for plotting contours for 3D data.

I've also removed some trailing whitespace that was introduced in PR #107.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Notebook/_plot.py
+++ b/python/BioSimSpace/Notebook/_plot.py
@@ -164,9 +164,13 @@ def plot(
         raise TypeError("'x' must be of type 'list'")
 
     else:
-        # Make sure all records are of the same type.
-        _type = type(x[0])
-        if not all(isinstance(xx, _type) for xx in x):
+        # Make sure all records are of the same type. Missing data will be
+        # None, so find the first unit type.
+        for idx, xx in enumerate(x):
+            if xx is not None:
+                _type = type(xx)
+                break
+        if not all(isinstance(xx, (_type, type(None))) for xx in x):
             raise TypeError("All 'x' data values must be of same type")
 
         # Convert int to float.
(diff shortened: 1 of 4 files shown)


chryswoods (dismissed):
This looks great.

PR #119, opened 2023-06-25
Bugfix Somd Analysis

fjclark:

This PR ensures that the temperature is read correctly from SOMD simfile.dat files when the temperature is given in Celsius. This fixes the issue with incorrect free energy changes [described here](https://github.com/OpenBioSim/BioSimSpaceTutorials/issues/12#issuecomment-1594937888). 

It also seemed like the MBAR error was being read from the wrong row of the dataframe.

## Suggested reviewers:
@lohedges

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
@@ -723,12 +723,22 @@ class Relative:
             found_temperature = False
             with open(file, "r") as f:
                 for line in f.readlines():
-                    t = None
+                    temp = None
                     start = "#Generating temperature is"
                     if start in line:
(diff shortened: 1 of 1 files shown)


lohedges:
Thanks for this. I'm not sure how this code originally ended up in the Sandpit. I think Exs might have just copied in some of Anna's work from her feature branch, so would have never got bugfixes. They also don't use SOMD, so might not have noticed it. 

@annamherz: Just tagging you in so that you are aware. I assume that your own version doesn't have this issue.

annamherz:
Hello! Actually this isn't in the feature branch, but I haven't used it to analyse anything where the temperature wasn't in K so it hadn't shown up as an issue. I'll copy it over so it's consistent there too.

lohedges (approved):

PR #121, opened 2023-06-26
Add support for Sire as a trajectory backend

### LINKED ISSUE #118: [BUG] Trajectory.getFrames returns system before updating original system
[This](https://github.com/OpenBioSim/biosimspace/blob/f2acfc00cb5eb9675822d6dd57f2f4c463969013/python/BioSimSpace/Trajectory/_trajectory.py#L201) return statement needs to be removed.

lohedges:
This PR adds support for Sire as a trajectory parsing backend within the `BioSimSpace.Trajectory` sub-package. This means that Sire can be used as a native trajectory backend for all engines that we currently support. At present, GROMACS is still parsed with MDAnalysis, since users have requested the use of the `tpr` format to preserve information that would otherwise be lost with, say gro, e.g. `bonds`, etc. This can obviously be achieved with Sire too, since it can parse any compatible format along with the trajectory, e.g. a top file. I'll port this over at a later date once I've given users some examples for how this works. (I don't want to break any external workflows that rely on this.)

The PR also closes #118.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/Trajectory/_trajectory.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/Trajectory/_trajectory.py
@@ -24,7 +24,7 @@
 __author__ = "Lester Hedges"
 __email__ = "lester.hedges@gmail.com"
 
-__all__ = ["getFrame", "Trajectory"]
+__all__ = ["getFrame", "Trajectory", "backends"]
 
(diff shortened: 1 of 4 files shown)


chryswoods (approved):
Looks good - it's nice that BioSimSpace is so modular that adding in a new trajectory backend isn't too painful

PR #122, opened 2023-06-26

lohedges:
This PR synchronises `devel` with the [Exscientia remote](https://github.com/Exscientia/biosimspace/tree/devel). This was quite a painful process as conflicts arose due to Exscientia internally sychronising outdated fixes from our `main` which had been superseded on `devel`, but doing so at a later date that the commits to `devel`. Because of other edits to the same file, this wasn't a simple case of overwriting with our changes.

Exscientia have added support for writing state information for AMBER and GROMACS to a parquet file using `pandas`. This means that most tests within their Sandpit involving `Process.Amber` and `Process.Gromacs` will now fail unless an appropriate parquet backend is installed. I note that _only_ the `pyarrow` backend works, i.e. the code is not agnostic to the backend. Testing with `fastparquet` will fail. It looks like Exscientia uses `pyarrow` in their workflows, so this likely isn't a problem, but is something that we would need to be aware of if moving the code out of the sandpit in an interoperable way. (Most errors were related to columns without a string for a label.) To ensure that the tests can run with-and-without `pyarrow` installed, I check for the backend using `pandas.io.parquet.get_engine("pyarrow")`, skipping tests if an exception is raised.

Certain sandpit tests now require `alchemlyb` version 2.1.0 and above. I've added a version test in the test files and an additional condition to the `skipif` for the affected tests. (We currently have a pin on `alchemlyb` due to requiring an older version of `pymbar` for the native SOMD analysis via `analyse-freenrg`. This can be removed once we merge in the new analysis code.)

I've moved all input files that are related _only_ to sandpit tests to the `tests/Sandpit/Exscientia/input` directory. This ensure that the sub-test suite is self-contained.

I've tested locally only on Linux, so will fix any test issues on other platforms based on the CI output, e.g. missing guards around tests.

## Checklist:

## Suggested reviewers:
@chryswoods


lohedges:
Ah, another annoying feature of the merge is that it's overwritten with the requirements file from main, so is looking for a sire version that doesn't exist with the `dev` label. Will fix and re-push.

lohedges:
Okay, it was an easy fix. They are using the following protocol:

```python
    protocol = Production(
        timestep=0.1 * femtosecond,
        runtime=0.1 * femtosecond,
        temperature=0 * kelvin,
    )
```

Because of the precision they are using when writing to file, this gives:

```
dt = 0.000
```

which, unsurprisingly, isn't allowed. Not sure why it works with 2022.4.

lohedges:
Just to note that I plan to submit another PR where I do some tidying, such as moving some boiler plate to a `conftest.py`

lohedges:
I've now refactored the tests to move checks for platform dependent packages to `conftest.py` files. I've also added a quick fix to handle changes to the `ipywidgets.FileUpload` API. This has been tested using `ipywidgets` 8.0.6 on macOS, and using the old API for `7.7.5` on Linux.

xiki-tempula:
> It looks like Exscientia uses pyarrow in their workflows, so this likely isn't a problem, but is something that we would need to be aware of if moving the code out of the sandpit in an interoperable way. (Most errors were related to columns without a string for a label.) To ensure that the tests can run with-and-without pyarrow installed, I check for the backend using pandas.io.parquet.get_engine("pyarrow"), skipping tests if an exception is raised.

pyarrow is a dependency of alchemlyb=2.1.0

lohedges:
Okay, thanks for clarifying. I'll still need the test guards since this won't be installed during our CI. It's also still possible to use the parquet functionality without touching any of the alchemlyb code.

chryswoods (approved):
Looks good - tested locally and looks ok.

lohedges:
The builds on `devel `fail because of a missing `2023.3.0.dev` Sire package. I'll be building against this until I've made the release, after which I'll update to `2023.4.0.dev`.

PR #125, opened 2023-07-03
Fix issue #124

### LINKED ISSUE #124: [BUG] Check for non-periodic Cartesian space in process setup
I'm not sure this wasn't caught by the local CI for the release, but we need to add a few more checks for a non-periodic space when setting up a process. (Sire now adds a default Cartesian space if no property is present.) This was fixed for NAMD [here](https://github.com/OpenBioSim/biosimspace/pull/120), but needs to be applied for _all_ engines.

lohedges:
This PR closes #124. We now check that the space property of the system is periodic for _all_ supported engines, to ensure that vacuum simulations are run correctly. Note that there is some code duplication here since not all process classes have been ported to the new configuration generator code. The GROMAC code is duplicated since we need to write a modified GRO file for vacuum simulations, which is done in the process setup, not the configuration generator.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/.github/workflows/devel.yaml
+++ b/.github/workflows/devel.yaml
@@ -13,7 +13,7 @@ jobs:
       max-parallel: 9
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11"]
         platform:
           - { name: "windows", os: "windows-latest", shell: "pwsh" }
           - { name: "linux", os: "ubuntu-latest", shell: "bash -l {0}" }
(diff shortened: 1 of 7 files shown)


lohedges:
Looks like I now need to use py39, 310, and 311. However, some packages are still missing, e.g. [this](https://anaconda.org/openbiosim/sire/files?version=2023.4.0.dev) doesn't contain a full set.

lohedges:
I've updated the Python versions but will wait to re-run the CI until the Sire packages are available.

chryswoods (approved):
All good - sorry that this was an annoying change. I think having `space` and `time` always defined will help us in the long term

lohedges:
No problem, it completely makes sense, I just have no idea why this only became an issue post-release, since I was building against a version of Sire where this change was already present :man_shrugging:

PR #128, opened 2023-07-07
Fix issues caught by BioSimSpaceTutorials

lohedges:
This PR fixes some issues that were caught by the [BioSimSpaceTutorials](https://github.com/OpenBioSim/BioSimSpaceTutorials) suite. (Why some of these weren't caught by the previous run through, I don't know.)

Changes are:

* Allow the user to specify a trajectory backend when calling `process.getTrajectory`. This is required since the default Sire backend can't (yet) compute the RMSD correctly. I've also updated the example notebook in `demos` to specify a backend.
* Fix the PLUMED version check required for determining whether and auxiliary file is needed to enable funnel metadynamics simulations.
* Pass the `process._work_dir` attribute to the `Process.Plumed` constructor as a string within `Process.OpenMM`. Previously, the `_Utils.WorkDir` object was being passed, so a type check was failing.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/demo/interactive_md.ipynb
+++ b/demo/interactive_md.ipynb
@@ -336,10 +336,10 @@
    "outputs": [],
    "source": [
     "# Search the system for a residue named ALA. Since there is a single match, we take the first result.\n",
-    "residue = system.search(\"resname ALA\")[0]\n",
+    "atoms = system.search(\"resname ALA\")\n",
     "\n",
     "# Get the indices of the atoms in the molecule, relative to the original system.\n",
-    "indices = [system.getIndex(x) for x in residue.getAtoms()]\n",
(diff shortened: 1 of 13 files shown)


chryswoods (approved):
Looks good - could I have some tests that show the difference in the rmsd calculated by sire. I am pretty sure the code is correct - it will be different definitions of reference or subsets of atoms being calculated I think. It would be good to debug to find a way to produce an equivalent rmsd.

lohedges:
Yes, no problem, I'll try to dig out the code that I wrote. It was off for a single frame and I tried with-and-without molecular wrapping. I imagine it will be very easy to fix once we work out why we're getting a difference.

Issue #132, opened 2023-07-10
[BUG] Incorrect alchemlyb extract function imported


lohedges:
During synchronisation with the Exscientia sandpit the wrong `alchemlyb` extract function used when guarding imports against missing packges, i.e.:

```python
from alchemlyb.parsing.gmx import extract as _extract
```

was changed to

```python
from alchemlyb.parsing.amber import extract as _extract
```

Issue #76, opened 2023-05-22
[BUG] rhombic dodecahedron triclinic spaces not working with OpenMM


lohedges:
The rhombic dodecahedron box parameters generated by Sire/BioSimSpace are not working with OpenMM. This is true both before and after the rounding fix introduces in OpenBioSim/sire#51. Internally OpenMM uses the same lattice reduction as us (it checks for reduced box dimensions), so I'm surprised this is not working. Other engines are fine with the parameters. Hopefully it's possible to apply a specific fix for this box type.

```python
In [1]: import BioSimSpace as BSS

In [2]: mol = BSS.Parameters.openff_unconstrained_2_0_0("C").getMolecule()

In [3]: box, angles = BSS.Box.rhombicDodecahedronHexagon(3*BSS.Units.Length.nanometer)

In [4]: solvated = BSS.Solvent.tip3p(mol, box=box, angles=angles)

In [5]: p = BSS.Protocol.Minimisation(steps=1000)

In [6]: process = BSS.Process.OpenMM(solvated, p)

In [7]: process.start()
Out[7]: BioSimSpace.Process.OpenMM(<BioSimSpace.System: nMolecules=622>, BioSimSpace.Protocol.Minimisation(steps=1000, restraint=None, force_constant=10.0 kcal_per_mol/angstrom**2), exe='/home/lester/.conda/envs/openbiosim/bin/sire_python', name='openmm', platform='CPU', work_dir='/tmp/tmpijqr1ske', seed=None)

In [8]: process.wait()

In [9]: process.isError()
Out[9]: True

In [10]: process.stderr()
/home/lester/.conda/envs/openbiosim/lib/python3.10/site-packages/numpy/core/getlimits.py:89: UserWarning: The value of the smallest subnormal for <class 'numpy.float32'> type is zero.
  return self._float_to_str(self.smallest_subnormal)
Traceback (most recent call last):
  File "/tmp/tmpijqr1ske/openmm_script.py", line 34, in <module>
    simulation = Simulation(prmtop.topology,
  File "/home/lester/.conda/envs/openbiosim/lib/python3.10/site-packages/openmm/app/simulation.py", line 103, in __init__
    self.context = mm.Context(self.system, self.integrator, platform, platformProperties)
  File "/home/lester/.conda/envs/openbiosim/lib/python3.10/site-packages/openmm/openmm.py", line 10069, in __init__
    _openmm.Context_swiginit(self, _openmm.new_Context(*args))
[...]

PR #135, opened 2023-07-11
Fix issue #134

### LINKED ISSUE #134: [BUG] ParmEd box vectors are occasionally incompatible with OpenMM
Occasionally triclinic box vectors generated by `parmed` (having read box dimensions and angles from AMBER rst7 format) are incompatible with `openmm`, which raises the following exception:

```python
Traceback (most recent call last):
  File "test1/run.py", line 146, in <module>
  File "test1/run.py", line 56, in test_openmm
  File ".../openmm/openmm.py", line 2292, in setPeriodicBoxVectors
    return _openmm.Context_setPeriodicBoxVectors(self, a, b, c)
openmm.OpenMMException: Periodic box vectors must be in reduced form.
```

This can be avoided by passing the `parmed` box vectors through the internal `openmm.app.internal.unitcell.reducePeriodicBoxVectors` function before passing them through to the context.

lohedges:
This PR closes #134. To work around rounding issues with triclinic box vectors following conversion from dimensions and angles we now perform a pre lattice reduction using the internal OpenMM functionality. This isn't needed when using the native OpenMM AMBER format parsers, so can be removed once the fix related to [this](https://github.com/OpenBioSim/biosimspace/issues/76) previous issue has made its way into a conda-forge release of OpenMM.

Note that a single test may fail for macOS py311. This is due to a precision issue, which has been adjusted in #132.

As before, the code duplication here is because the OpenMM protocols are currently (mostly) independent as we work out the best options for each. This will be refactored once we port to `BioSimSpace._Config`. (We might just use the native Sire conversion, so do away with this entirely.)

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Process/_openmm.py
+++ b/python/BioSimSpace/Process/_openmm.py
@@ -345,7 +345,10 @@ class OpenMM(_process.Process):
                 self.addToConfig("simulation.context.setPositions(prm.positions)")
             self.addToConfig("if prm.box_vectors is not None:")
             self.addToConfig(
-                "    simulation.context.setPeriodicBoxVectors(*prm.box_vectors)"
+                "    box_vectors = reducePeriodicBoxVectors(prm.box_vectors)"
+            )
+            self.addToConfig(
+                "    simulation.context.setPeriodicBoxVectors(*box_vectors)"
             )
             self.addToConfig(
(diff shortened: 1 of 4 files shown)


chryswoods (approved):
👍

PR #146, opened 2023-08-01
Fix issue #145

### LINKED ISSUE #145: [BUG] Incorrect use totalSteps in OpenMM StateDataReporter
We need to pass in the total number of simulation steps, not a boolean. This is needed so that OpenMM can correctly report the remaining simulation time.

lohedges:
This PR closes #145 and closes #145. It fixes the use of the `totalSteps` parameter within the OpenMM `StateDataReporter`. Previously a boolean was passed (which was silently ignored). We actually need to pass an integer that specifies the total number of simulation steps. This allows the reporter to correctly report things like the remaining simulation time, or the progress percentage.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Process/_openmm.py
+++ b/python/BioSimSpace/Process/_openmm.py
@@ -2083,6 +2083,11 @@ class OpenMM(_process.Process):
             is_time = True
             is_temperature = True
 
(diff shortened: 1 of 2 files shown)


chryswoods (approved):
👍

PR #149, opened 2023-08-02
Fix issue #148

### LINKED ISSUE #148: [BUG] Water molecule coordinates are setting to (0,0,0) (reposting)
I am re-posting issue [427](https://github.com/michellab/BioSimSpace/issues/427) here for future discussion as @lohedges requested.

lohedges:
This PR closes #148 and closes michellab/BioSimSpace#427. The code has been updated to correctly update the water topology to match the GROMACS naming convention in _both_ the Gro87 and GroTop files. Since a user may wish to preserve a particular naming convention, e.g. for crystal waters, I've added an option to disable the topology swap if desired, e.g. if they are just writing to file. A few optimisations have been added following the addition of this parameter, since it means that we can avoid double checking the topology in cases where we know that it has already been swapped, e.g. during the setup of a simulation with a particular engine.

I've not resolved the issue of the Gro87 file being parsed incorrectly. I'm not sure what we can do about this (in a general way, at least) since the file is valid (doesn't trip the existing error handling), but is formatted incorrectly.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/IO/_io.py
+++ b/python/BioSimSpace/IO/_io.py
@@ -729,12 +729,14 @@ def saveMolecules(filebase, system, fileformat, property_map={}, **kwargs):
                 _os.rename(file, new_file)
                 file = [new_file]
             elif format.upper() == "GRO87":
+                system_copy = system.copy()
+                system_copy._set_water_topology("GROMACS", _property_map)
                 # Write to 3dp by default, unless greater precision is
                 # requested by the user.
                 if "precision" not in _property_map:
                     _property_map["precision"] = _SireBase.wrap(3)
                 file = _SireIO.MoleculeParser.save(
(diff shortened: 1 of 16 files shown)


lohedges:
It looks like the `match_water` option won't work correctly for the case of writing to GroTop format when different naming schemes are used for the _same_ water topology within a system, e.g. if you have some crystal waters along with regular waters. The GroTop parser deduplicates the molecule types, so you'll only end up with a single type in the top file.

lohedges:
This has been consumed into a larger crystal water related feature that I'm working on so I'll close this when I submit that PR. I'll still pull out the specific fix implemented here for a backport, i.e. forgetting to update the water topology for Gro87 format on write.

chryswoods (approved):
👍

PR #154, opened 2023-08-09
Add support for crystal waters

### LINKED ISSUE #67: [BUG] Adding a cofactor
Hi,

I have tried to pull all togheter a protein a ligand a and B and a cofactor. I managed to run the parametrization of the four components. Also before to merge all togheter I create the mapping between the two ligands (A and B), then I merged then after that using the next line complx = merged + RNMT + cofactor.

The I tried to minimize the system but it fails to do that. The error is: RuntimeError: 'gmx genion' failed to add ions! Perhaps your box is too small?. I increase the size of the box, but it fails no matter the size. When I merge only protein and merged ligands, it works well. 

Thanks a lot

Cesar

### LINKED ISSUE #112: [BUG] Crystallographic waters could be removed during solvation
I came across this potential problem when reading [this](https://github.com/openmm/openmm/issues/4103) issue thread. For OpenMM, existing crystallographic waters are (now) correctly preserved when adding ions. There is no logic do deal with this when using `gmx genion`, other than specifying the minimum distance between ions and non-solvent with `-rmin`, which doesn't seem to work very well. Although I imagine that this issue would rarely occur, we might need to add some logic to check which waters were replaced, and re-run if any crystallographic waters were removed. (Alternatively, we could just switch to using `openmm.Modeller` if it's more reliable.)

### LINKED ISSUE #141: [BUG] genion fails with crystallographic water in a protein-ligand system
**Describe the bug**
In my system, I have a protein with two crystallographic waters bound to a ligand.
BSS.Solvent.solvate() fails with the following error:
image

Looking into the genions.err, I found the error message:
image

And this is because in the input system, I have in order:
- protein 
- crystallographic waters
- ligand

Therefore, when solvent waters are added, genions sees water molecules that are not continuous but with a ligand in between.
How to best deal with this?

**To Reproduce**
```
import BioSimSpace as BSS

protein_coord = BSS.IO.readMolecules('protein.pdb')[0]
ligand_coord = BSS.IO.readMolecules('ligand.sdf')[0]

protein = BSS.Parameters.parameterise(protein_coord , 'ff14SB', water_model='tip3p').getMolecule()
ligand = BSS.Parameters.parameterise(ligand_coord, 'gaff2').getMolecule()
system = protein + ligand

box = [100 * BSS.Units.Length.angstrom] * 3
angles = [100 * BSS.Units.Angle.degree] * 3
solvated = BSS.Solvent.solvate('tip3p', system, box=box, angles=angles, work_dir='tmp')
```

**Expected behavior**
BSS.Solvent.solvate() should not fail

 - Ubuntu
 - Version of Python: 3.10.12
 - Version of BioSimSpace: 2023.3.0
 - I confirm that I have checked this bug still exists in the latest released version of BioSimSpace: [yes/no] no (not in the dev version)

### LINKED ISSUE #148: [BUG] Water molecule coordinates are setting to (0,0,0) (reposting)

lohedges:
This PR adds support for setting up simulations for systems containing crystal waters. While we have always handled existing water molecules for solvation, this update provides an easy way for users to load PDB systems with embedded crystal waters, then parameterise and solvate them. The waters need not be unique molecules, since they can be extracted using the `.extract` method, and also do not need to have hydrogen atoms. For the latter, I've added a `ensure_compatible` option to the parametise function to allow users to choose when to ignore the `makeCompatibleWith` check, e.g. when `tLEaP` adds the missing hydrogens when parameterising the solvent. (The default is `ensure_compatible=True`.) We now also make sure that `gmx genion` doesn't accidentally remove any crystal waters when adding ions during solvation. A full tutorial can be found [here](https://github.com/OpenBioSim/biosimspace/blob/feature_crystal_water/doc/source/tutorials/crystal_water.rst).

Closes michellab/BioSimSpace#32
Closes michellab/BioSimSpace#427
Closes #67
Closes #112
Closes #141
Closes #148
Closes #149

## Suggested reviewers:
@chryswoods

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -16,6 +16,9 @@ Steps to reproduce the behavior:
 2. Run the code '...'
 3. This is the exception that was raised / this is what went wrong.
 
+(If possible, please quote code snippets using the markdown formatting
+described [here](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code). This makes it easy to copy and paste the example, avoiding any translation errors.)
+
 **Expected behavior**
 A clear and concise description of what you expected to happen
 (or, it should not raise an exception if an exception was raised)
--- a/doc/source/tutorials/crystal_water.rst
+++ b/doc/source/tutorials/crystal_water.rst
@@ -1,0 +1,135 @@
+.. _ref_crystal_water:
+=============
+Crystal water
+=============
+
+.. toctree::
+   :maxdepth: 2
+
+Water molecules are important for the structure, function, and dynamics of
+proteins. When setting up a simulation it may be important to preserve
+specific water molecules, often referred to as *crystal waters*. For example,
+many structures from the `Protein Data Bank <https://www.rcsb.org/>`__
+contain the coordinates of water molecules that are resolvable via X-ray
+crystallography. In this example, we will use BioSimSpace to parameterise
+and solvate a `tyrosine kinase 2 <https://en.wikipedia.org/wiki/Tyrosine_kinase_2>`__
+protein structure containing two crystal waters within its binding site.
+
+Firstly, let's load the PDB structure:
+
+>>> import BioSimSpace as BSS
+>>> tyk2_xtal = BSS.IO.readMolecules(
+...    BSS.IO.expand(
+...        BSS.IO.tutorialUrl(), "tyk2_xtal.pdb"
+...    )
+...)[0]
+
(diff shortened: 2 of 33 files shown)


lohedges:
Looks like a few bits of the sandpit will need to be updated to pass parameters using keyword argument format. Will fix this now.

lohedges:
Hi @annamherz, I was wondering if it would make sense for you to take a look at this feature, since you shared the original example file for me. @jmichel80 mentioned that you might also be looking at similar systems in your upcoming placement.

chryswoods (approved):
👍

PR #155, opened 2023-08-11
Add unified FEP analysis using alchemlyb

lohedges:
This PR ports @annamherz's FEP analysis from the [feature-analysis](https://github.com/michellab/BioSimSpace/tree/feature-analysis) branch of the legacy michellab repo. This is a necessary first step to allow the porting of AMBER FEP and ABFE from the Exscientia sandpit.

In comparison with the original michellab branch there are a few notable changes:

* The estimator is no longer an option to the `FreeEnergy.Relative` constructor. Since the analysis is entirely independent from the simulation and can be run statically on existing output there is no need for the estimator to be specified when an instance is created. (This would only be needed if the estimator changed the actual output of the simulation.)
* I've refactored to reduce code duplication and ease future maintenance. For example, there is now a single `_somd_extract` method that takes the estimator as a parameter. Combining this with `itertools.partial`, it's possible to create a callable with the expected signature in the function from which it is called.
* I've removed the plotting functionality since the overlap can already be shown using the [BioSimSpace.Notebook.plotOverlap](https://github.com/OpenBioSim/biosimspace/blob/8d11aec199aa2c309b6306ad34616b479b9243fe/python/BioSimSpace/Notebook/_plot.py#L538) function. In doing so, I decided to always return `None` for the overlap when the chosen estimator (or method) wouldn't return one, rather than returning something different for TI. Returning different data types, or adding support for more plotting functionality is something that we can consider at a later date.
* I've added a [unit test](https://github.com/OpenBioSim/biosimspace/blob/68a13d49a4e4db88d0c246b9b35faaeba340025c/tests/FreeEnergy/test_relative.py#L52-L95) that runs analysis against some existing output data to ensure that the results are as expected. Thanks to @annamherz for testing using your local branch to ensure that we were both getting the same numbers.

Once this is approved and merged, please direct any future development work, e.g. bug-fixes, improvements, to this repo.

Closes michellab/BioSimSpace#365.

## Checklist:

* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [y] (Already documented, this just changes the backend used by the existing API.)

## Suggested reviewers:
@annamherz

### DIFF
--- a/doc/source/api/index_Box.rst
+++ b/doc/source/api/index_Box.rst
@@ -11,11 +11,7 @@ To support triclinic boxes that work across the range of molecular simulation
 engines that support, we represent the triclinic space in reduced form, using
 the approach documented in Appendix A of Chapter 3 from
 `"Molecular dynamics of sense and sensibility in processing and analysis of data" <https://research.rug.nl/files/2839528/01_c1.pdf>`__
-by Tsjerk A. Wassenaar. Due to the fixed-width format used to represent box
-dimensions and angles in the various molecular input files, repeated reading
-and writing can lead to oscillation of the box angles on reduction due to
-rounding precision errors. To account for this, we add a small bias so that
-we always round in a consistent direction.
+by Tsjerk A. Wassenaar.
 
 .. automodule:: BioSimSpace.Box
 
--- a/doc/source/tutorials/crystal_water.rst
+++ b/doc/source/tutorials/crystal_water.rst
@@ -1,0 +1,135 @@
+.. _ref_crystal_water:
+=============
+Crystal water
+=============
+
+.. toctree::
+   :maxdepth: 2
+
+Water molecules are important for the structure, function, and dynamics of
+proteins. When setting up a simulation it may be important to preserve
+specific water molecules, often referred to as *crystal waters*. For example,
+many structures from the `Protein Data Bank <https://www.rcsb.org/>`__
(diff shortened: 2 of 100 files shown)


lohedges:
This seems to repeatedly fail on Python 3.9 (with the same dependency versions). I'll need to debug locally. The windows tests appear to fail because the glob logic for working out the list of output files is failing, presumably due to the use of a forward slash in the split command. I'll see if this can be made OS agnostic.

lohedges:
Phew, thankfully both issues were easy to fix. All appears to be working now.

annamherz:
Hello! Would it be possible to keep a function similar to get_u_nk as in [this recent commit](https://github.com/michellab/BioSimSpace/commit/a6933302f545b83ce72b595d11bdad84c29852ab), as for some more advanced analysis I need to be able to get the extracted raw df? This is probably more cumbersome in the code but I think could potentially be useful for future code changes as well?

lohedges:
Yes, no problem, I'll add that this afternoon.

annamherz (approved):
Hello, sorry for the delay! I added some tiny fixes but these needed to be opened in a PR (https://github.com/OpenBioSim/biosimspace/pull/182)

lohedges:
Thanks, @annamherz, much appreciated. The fixes look good, will merge them in when I'm at the computer. That was the correct way to submit, i.e. fork from my branch then PR against it.

PR #157, opened 2023-08-21
Fix handling of connectivity changes during molecular perturbations

lohedges:
This PR makes some fixes associated with the _connectivity_ of merged molecules:

* In the `BioSimSpace.IO.readPerturbableSystem` function we now set a _single_ `connectivity` property if `connectivity0` and `connectivity1` of the two loaded end states are the same. This means that the resulting system is consistent with the _original_ perturbable system, and what would be obtained when streaming to/from a `.bss` file.
* Merges that _do_ change connectivity are now checked for in the correct way, i.e. by creating a unique connectivity object for both end states and seeing how the connectivity changes between them. (Previously this was done using a single connectivity for both end states.) If the connectivity _does_ change between end states, then we do attach two properties to the molecule, i.e. `connectivity0` and `connectivity1`.
* The logic for the `intrascale` matrix generation has been updated to handle different connectivities. This is the exact same approach as already used in the Exscientia Sandpit for their "region of interest" merge, which can be used for mutations. (I'm not yet sure if this is correct in all cases, but it is certainly more correct than what we have now.)

(I've not ported the updates to the merge code to the Exscientia sandpit since they already have their own functionality for handling perturbations of this nature.)

## Checklist

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Align/_merge.py
+++ b/python/BioSimSpace/Align/_merge.py
@@ -927,16 +927,19 @@ def merge(
                 "or 'allow_ring_size_change' options."
             )
 
-    # Create the connectivity object
-    conn = _SireMol.Connectivity(edit_mol.info()).edit()
+    # Create the connectivity objects.
+    conn0 = _SireMol.Connectivity(edit_mol.info()).edit()
+    conn1 = _SireMol.Connectivity(edit_mol.info()).edit()
 
-    # Connect the bonded atoms. Connectivity is the same at lambda = 0
-    # and lambda = 1.
+    # Connect the bonded atoms in both end states.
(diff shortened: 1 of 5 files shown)


chryswoods (approved):
Thanks - I like this. This is a good first step towards the `connectivity` property becoming the single source of truth of how we should treat the connectivity of the merged molecule. You could imagine for perturbations involving changed bonding that `connectivity` could become a merged object, i.e. having connections if atoms are bonded in either end state. This would be much like the atom layout having atoms if they are non-dummy in either end state. The perturbation layer could then build an initial intrascale from the merged `connectivity`, and work out deltas from the difference of this and the end states.

PR #163, opened 2023-09-01
Fix issue #162

### LINKED ISSUE #162: [BUG] SOMD ABFE tests are incorrectly skipped when GROMACS isn't installed
This can be resolved by removing the offending skipif decorator [here](https://github.com/OpenBioSim/biosimspace/blob/6fdda73107ce873d6ac04d6c9b30185b23ea2996/tests/Sandpit/Exscientia/FreeEnergy/test_alchemical_free_energy.py#L200).

lohedges:
This PR closes #162. If this fails on macOS, which is the case for the same code in the Exscientia sandpit, then I'll need to debug locally to resolve.


lohedges:
Closing since GROMACS is needed. A solvation was performed within the test setup, which I hadn't spotted.

PR #164, opened 2023-09-01
Ensure that SOMD ABFE pert files are always correct

fjclark:
This fixes issues with the writing of SOMD pert files for ABFE calculations where "atom" could be missed from the start of the atom records and "endmolecule" was always absent. I've extended the tests to cover these sections of the pert files.

## Suggested reviewers:
@lohedges


lohedges:
Thanks for this. The test failure is due to a precision issue for restraints in the Exscientia sandpit. These continually seem to cause issues so I'll need to reduce the tolerance somehow.

lohedges (approved):

PR #166, opened 2023-09-08
Fix user commands that reference mol error:

dickensn:
Code caused "Argument #1 is of type String must be of type: [unit]" error because 'mol =' was being added to leap.txt after the user commands.  So any user commands that referenced it failed (e.g. solvateBox).

## If this is to fix a bug...

This pull request fixes issue Not a recorded issue I could find.

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): [y/n] y
* I confirm that I have permission to release this code under the GPL3 license: [y/n] y

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): [y/n] y
* I confirm that I have added a test for any new functionality in this pull request: [y/n] y
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [y/n] y
* I confirm that I have permission to release this code under the GPL3 license: [y/n] y

## Suggested reviewers:
@lohedges, @chryswoods

(you can also attach files by dragging & dropping)

### DIFF
--- a/python/BioSimSpace/Parameters/_Protocol/_amber.py
+++ b/python/BioSimSpace/Parameters/_Protocol/_amber.py
@@ -460,11 +460,13 @@ class AmberProtein(_protocol.Protocol):
                     file.write("source leaprc.water.tip4pew\n")
                 else:
                     file.write("source leaprc.water.%s\n" % self._water_model)
+            # If user commands reference mol, this needs to be first
+            file.write("mol = loadPdb leap.pdb\n")
             # Write extra user commands.
             if self._leap_commands is not None:
                 for command in self._leap_commands:
                     file.write("%s\n" % command)
-            file.write("mol = loadPdb leap.pdb\n")
+            # file.write("mol = loadPdb leap.pdb\n")
             # Add any disulphide bond records.
             for bond in disulphide_bonds:
                 file.write("%s\n" % bond)


lohedges:
Thanks, @dickensn. Unfortunately it's not as simple as this since the commands might need to come first depending on what you're doing, e.g. adding custom force field parameters. (This is the reason why it was originally added.) I think what would be best is if I added an option to control the point at which the commands are injected, e.g. `pre_mol` or `post_mol`. Hopefully that would cover most common use cases. I'll have a play around next week and let you know how I get on.

If you have a simple test failure for your issue then I can use that to debug and add to our unit testing suite. As long as it's not molecule specific, just the list of additional commands you are trying to add should be good.

Cheers.

lohedges:
It might also be possible to add an option for the user to view and edit the `LEaP` script before running, which might be more flexible.

lohedges:
I've created a feature branch to deal with this [here](https://github.com/OpenBioSim/biosimspace/tree/feature_167). This decouples the injection of custom force field parameters from additional `LEaP` commands, which might need to reference the molecule that is loaded. Hopefully this will handle most use cases. (For anything super custom, the user will probably just interface with `LEaP` directly.)

PR #170, opened 2023-09-18
Decouple custom parameters and additional LEaP commands

### LINKED ISSUE #167: Allow user commands to reference molecule in LEaP script
Currently the user defined `leap_commands` that are injected into the `LEaP` script are inserted _prior_ to the loading of the molecule. The original request for this feature was to allow the use of custom parameters, hence the commands needed to come _after_ the default parameters, but _before_ the molecule has been loaded. However, this approach means that the user cannot add any commands that make reference to the molecule itself, since it's not yet defined, e.g. see #166.

Unlike for `BioSimSpace.Process`, it is a bit more tricky to let the user see the generated `LEaP` script and modify it prior to execution. (This would be the most flexible solution.) As an alternative, I propose having _two_ user options: `custom_parameters` and `leap_commands`. The former is a list of paths to custom parameter files that should be loaded, the latter is additional commands, that may make reference to the molecule. This would allow us to place things in the right place in both cases and hopefully serve most use cases. (The flexibility of `leap_commands` is still limited, though.)

lohedges:
This PR closes #166 and closes #167.

* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [y/n]

## Suggested reviewers:
@chryswoods 

@AdeleHardie: Just tagging you in since the original implementation came from a feature request from yourself [here](michellab/BioSimSpace#182). Hopefully this still works for you. Let me know if you have any comments/suggestions.

### DIFF
--- a/python/BioSimSpace/Parameters/_Protocol/_amber.py
+++ b/python/BioSimSpace/Parameters/_Protocol/_amber.py
@@ -122,6 +122,7 @@ class AmberProtein(_protocol.Protocol):
         tolerance=1.2,
         max_distance=_Length(6, "A"),
         water_model=None,
+        custom_parameters=None,
         leap_commands=None,
         bonds=None,
         property_map={},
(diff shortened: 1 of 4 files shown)


lohedges:
(Just killed the CI since I'll need to merge other PR's into `devel` first, then merge back into this, i.e. so it has the updated angle parameters in the tests.)

chryswoods (approved):
All looks sensible. Documentation is clear :-)

PR #173, opened 2023-09-20
Fix issue #172

### LINKED ISSUE #172: [BUG] getSystem might replace space with an infinite Cartesian box
Sire now adds a default space on load if one isn't present, i.e. one of type`sire.vol.Cartesian()`. In the `getSystem` methods of `BioSimSpace.Process` objects, the space property of the _original_ system will be replaced with the one loaded from the trajectory or restart file, e.g. if the box changes during an NPT simulation. However, if no box is present in the file, then the original will now be replaced by a Cartesian space, rather then being preserved. This is the case for `BioSimSpace.Process.Somd`, but may also be an issue for other process classes too.

The solution is to check that the box is periodic before copying, e.g. using `space.isPeriodic()`.

lohedges:
This PR closes #172. We now check that the space of any loaded system is periodic before updating the property in the original system, e.g. when calling `getSystem()` on a process, or when parsing trajectory frames with reference to an existing system. This avoids the space being replaced by a `sire.legacy.Vol.Cartesian` space if the file format used to load the new system doesn't contain space information, which is the case for SOMD.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Process/_amber.py
+++ b/python/BioSimSpace/Process/_amber.py
@@ -442,9 +442,10 @@ class Amber(_process.Process):
             # Update the box information in the original system.
             if "space" in new_system._sire_object.propertyKeys():
                 box = new_system._sire_object.property("space")
-                old_system._sire_object.setProperty(
-                    self._property_map.get("space", "space"), box
-                )
+                if box.isPeriodic():
+                    old_system._sire_object.setProperty(
+                        self._property_map.get("space", "space"), box
+                    )
 
             return old_system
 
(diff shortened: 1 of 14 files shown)


lohedges:
Note that this does mean that `getSystem` is effectively broken for SOMD with NPT simulations, i.e. the space won't be updated correctly. This issue is because of the reordering of molecules in the system, meaning that I use a special restart PDB file to provide a way to remap to the original system based on a unique naming and indexing convention. I guess I might need to load the DCD file too in order to get box information. Will play around with this.

chryswoods (approved):
All good - sorry this was a pain caused by adding in the default space.

PR #175, opened 2023-09-22
Feature reduce

lohedges:
This PR exposes functionality related to the changes to `sire.legacy.Vol.TriclinicBox`, allowing the user to manually rotate the space and perform a lattice reduction. Keywords have also been added to `BioSimSpace.IO.readMolecules` to allow the user to perform these operations on load.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/doc/source/api/index_Box.rst
+++ b/doc/source/api/index_Box.rst
@@ -11,11 +11,7 @@ To support triclinic boxes that work across the range of molecular simulation
 engines that support, we represent the triclinic space in reduced form, using
 the approach documented in Appendix A of Chapter 3 from
 `"Molecular dynamics of sense and sensibility in processing and analysis of data" <https://research.rug.nl/files/2839528/01_c1.pdf>`__
-by Tsjerk A. Wassenaar. Due to the fixed-width format used to represent box
-dimensions and angles in the various molecular input files, repeated reading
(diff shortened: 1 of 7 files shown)


chryswoods (dismissed):
Looks good. I think it is good that we assume the perturbable properties are called xxx0 and xxx1 (as you've done to find the perturbable molecules coordinates and velocities properties). The user has freedom to use other names, but it makes the code more difficult. I've said this is the format in the new sire docs for merged molecules. It may be worth separately creating a "detailed guide" page on the BioSimSpace website describing the merged molecule layout and format.

lohedges:
Yes, that's a good point. I allow the user to pass a map when performing the merge (if the system has different names) but then it just uses what it finds and appends 0 and 1.

lohedges:
Hmm, apparently the updated test angles weren't on this branch, despite me merging across from `devel`. Will see what went wrong.

lohedges:
Nope, the change is there and it's built against the most recent development version of Sire. Will debug locally. I can't see that the recent Sire merge has somehow reverted the triclinic update?

lohedges:
(No, not possible since other tests would fail here.)

lohedges:
Ah, the issue is that the test _works_ if you run it in isolation (as I was doing) but not if you run them all. Presumably the space is being modified by another test. Will fix and repush.

PR #183, opened 2023-10-12

lohedges:
This PR synchronises devel with the [Exscientia remote](https://github.com/Exscientia/biosimspace/tree/devel).

## Suggested reviewers:
@chryswoods

### DIFF
--- a/.github/workflows/Sandpit_exs.yml
+++ b/.github/workflows/Sandpit_exs.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macOS-latest",]
-        python-version: ["3.8", ]
+        python-version: ["3.10",]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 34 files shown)


.github/workflows/Sandpit_exs.yml
```diff
@@ -35,12 +35,10 @@ jobs:
 
       - name: Install dependency
         run: |
-          mamba install -c conda-forge -c openbiosim/label/main biosimspace python=3.8 ambertools gromacs "sire=2023.2.2" alchemlyb pytest pyarrow "typing-extensions!=4.6.0" openff-interchange pint=0.21 "rdkit<2023" "jaxlib>0.3.7" tqdm
+          mamba install -c conda-forge -c openbiosim/label/main biosimspace python=3.10 ambertools gromacs "sire=2023.3" "alchemlyb>=2.1" pytest openff-interchange pint=0.21 rdkit "jaxlib>0.3.7" tqdm
```
chryswoods, line 38: Is this the right version for sire? Should it be `2023.3.2` if they want to pin to the last major release, or do they want `2023.4.0`?

chryswoods (approved):
Code looks good, subject to the query about the sire version.

The test failure for Linux looks weird. Why would OpenMM move the molecule only for that platform?

I'm happy to approve - good luck with the release :-)

.github/workflows/Sandpit_exs.yml
lohedges: Good point. I'll check what's on their current branch.

lohedges:
The CI script is exactly what is in their fork. @xiki-tempula: Just checking this is what you want. As for the OpenMM failures, I think these are the position restraint issues that I was talking about. The merge must have overwritten my edits.

lohedges:
There is also a failure due to be guarding on Windows which is correct in our `devel`. I don't really like this way of synchronisation since I have to assume that the Exs fork is the "truth", which means that fixes that I've made are sometimes overwritten. (I can't preferentially take my edits.) The correct method is for me to PR to them, but no-one seems to want this approach.

chryswoods (approved):
👍

xiki-tempula:
Thanks for the PR.
Yes, we are using Sire 2023.3.2 now but after the sync, we would like to go to Sire 2023.4.
@janjoswig Jan do you mind have to look to see if the position restraint is still fine after the sync?

janjoswig:
Sorry for being late here. I noticed the issue as well that the position restraints apparently are not working in the same way depending on the platform and it still puzzles me. I just confirmed that the code is still working locally on MacOS and Linux and that atoms move without restraints and stay put with restraints. The reason for the different tolerances for each protocol is basically that these were the necessary thresholds to make the pipeline pass as well.

lohedges:
No problem. I'm pleased it's something that you've noticed too. I haven't had time to debug it fully and it's not completely reproducible anyway (it only fails on the one platform intermittently). As such, I've just adjusted the tolerance for that one particular test.

PR #190, opened 2023-10-23
Pass match_water parameter to specialised solvation functions

### LINKED ISSUE #185: [BUG] match_water parameter not passed through to solvation function
When calling the generic `BioSimSpace.Solvent.solvate` function, the `match_water` parameter isn't passed through to the specialised water model function.

lohedges:
This PR fixes a bug where the `match_water` parameter wasn't passed from the generic `BioSimSpace.Solvent.solvate` function through to the specialised functions for the supported water models. I've generalised the test so it uses both function types so that we catch similarly missed parameters in future, e.g. when anything else is added.

Closes #185.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/Solvent/_solvent.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/Solvent/_solvent.py
@@ -142,6 +142,7 @@ def solvate(
         ion_conc,
         is_neutral,
         is_aligned,
+        match_water,
         work_dir,
         property_map,
     )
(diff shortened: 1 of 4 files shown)


chryswoods (commented):
Looks good - Am I right in remembering that the root cause of the bug was the change in the number of positional arguments, so `match_water` wasn't passed correctly? Is it worth switching to keyword arguments for these many-argument functions?

(no need to do this for this PR fix - but something to consider for the future)

lohedges:
Yes, that's right. They should really be keywords. They are everywhere other than the validation function, so I'll update that in future. There used to only be a few, but it makes sense to switch to keywords now they have proliferated.

No worries. I forgot to remove the review request since it was identical to the previous PR.

PR #197, opened 2023-11-07
Fix issue #196

### LINKED ISSUE #196: [BUG] Velocity removal check is invalid if molecule has perturbable and non-perturbable properties
We currently remove velocity properties when not _all_ molecules in the system have them, e.g. when combining molecules. The check makes sure that the total number of velocity properties (non-perturbable and perturbable) is not-equal to the number of molecules in the system. However, it is possible for a molecule to have both types of property, so we should instead check that the total number is not less than the number of molecules.

lohedges:
This pull request closes #196 by checking for perturbable _and_ non-perturbable velocity properties and making sure that we don't double count. I've also added a test relevant to OpenBioSim/sire#123.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/doc/source/tutorials/crystal_water.rst
+++ b/doc/source/tutorials/crystal_water.rst
@@ -52,7 +52,7 @@ protein:
 
 And then the water:
 
->>> xtal_water = BSS.Parameters.tip3p(
+>>> xtal_water = BSS.Parameters.ff14SB(
 ...     xtal_water,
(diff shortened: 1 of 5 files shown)


lohedges:
Note that test failures are expected until the Sire PR is merged.

chryswoods (approved):
Looks good - thanks for adding a good test

PR #203, opened 2023-11-13
Internal updates due to Sire API fixes

lohedges:
This PR applies some internal updates due to Sire changes introduced via OpenBioSim/sire#119 and OpenBioSim/sire#120. For the `cursor.rotate` update I've chosen (for now) to use `rotate_velocities=False` and manually rotate the velocities using the property map, i.e. adjusting what `coordinates` is mapped to. This is so the user can rename things how they like. I also manually apply a rotation for both end states when perturbable molecules are present. By adding`rotate_velocities=False` I'm simply making sure that I don't perform a double rotation. (Note that I could use the end state linking functionality, but this is simpler for now.)

## Suggested reviewers:
@chryswoods


chryswoods (approved):
👍

PR #204, opened 2023-11-13
Bug Fix: Avoid inserting pair into boresch_dof_data until after try-except block

fjclark:
This fixes a bug in the restraint search algorithm for Boresch restraints. In the event that the other Boresch anchor points could not be generated for a given pair of atoms, previously the pair was inserted into `boresch_dof_data` anyway, with an empty dictionary for its value. I've shifted the insertion of the pair into the dictionary to after the try-except block, which avoids inserting failed pairs.

## Suggested reviewers:
@lohedges, @chryswoods


lohedges (approved):
Thanks for this. Don't worry about the CI failure, that was due to a numerical precision issue in one of the Exs tests that happens from time-to-time on Python 3.11.

PR #210, opened 2023-11-22
Fix #208

### LINKED ISSUE #208: [BUG] BioSImSpace._SireWrappers.Molecule.extract changes MolNum
The option to preserve the `MolNum` when calling the `.extract` method on a `Molecule` doesn't appear to work correctly. (It is `True` by default.)

```python

In [2]: m = BSS.IO.readMolecules(BSS.IO.expand(BSS.tutorialUrl(), ["ala.top", "ala.crd"]))[0]
Downloading from 'https://biosimspace.openbiosim.org/m/ala.top'...
Unzipping '/tmp/tmp8mfv1juu/ala.top.bz2'...
Downloading from 'https://biosimspace.openbiosim.org/m/ala.crd'...
Unzipping '/tmp/tmp8mfv1juu/ala.crd.bz2'...

In [3]: m.number()
Out[3]: 6

In [4]: idxs = [atom.index() for atom in m.search("not resname ALA").atoms()]

In [5]: mm = m.extract(idxs, renumber=False)

In [6]: mm.number()
Out[6]: 633
```

### LINKED ISSUE #213: [BUG] The intrascale matrix read by the AmberPrm parser disagrees with the saved one for perturbable systems
I'm not yet sure if this a general issue with the `sire.legacy.IO.AmberPrm` parser, or is a specific problem for perturbable systems, but the `intrascale` matrix loaded via `readPerturbableMolecules` doesn't agree with the one of the perturbable molecule that was saved in either end state. This is causing issues when running simulations with `somd2`. For now, a workaround is to regenerate the matrix using the `sire.legacy.IO.GroTop` parser.

lohedges:
This PR closes #208  by fixing the `renumber` option in `BioSImSpace._SireWrappers.Molecule.extract`. The idea behind this was to allow the user to extract a partial molecule then update the same original molecule in a system in one step, rather than needing to remove the old molecule and add the new one, which would (potentially) change the order. (There are ways to preserve this, i.e. you can update by index.)

Since this option was already broken, and hasn't seemed to cause any issue, perhaps it's better to remove it entirely and just deal with a different number, e.g. updating by index as suggested above. (There is code to do this and preserve the existing molecular order, i.e. you just replace the molecule in its existing position in the system.)

I've also added a quick fix that closes #213 by using the `sire.legacy.IO.GroTop` parser to reconstruct the intrascale matrices for the end states when loading a perturbable system from file. This has been tested via `somd2`.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -20,6 +20,8 @@ Devel
 * Minor internal updates due to Sire API fixes (`#203 <https://github.com/OpenBioSim/biosimspace/pull/203>`__).
 * Fix bug in the BSS Boresch restraint search code (`@fjclark <https://github.com/fjclark>`_) (`#204 <https://github.com/OpenBioSim/biosimspace/pull/204>`__).
 * Added SOMD and GROMACS support for multiple distance restraints for ABFE calculations (`#178 <https://github.com/OpenBioSim/biosimspace/pull/178>`__).
+* Fix ``renumber`` option in :meth:`extract <BioSimSpace._SireWrappers.Molecule.extract>` method (`#210 <https://github.com/OpenBioSim/biosimspace/pull/210>`__).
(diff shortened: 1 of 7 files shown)


chryswoods (approved):
👍

PR #224, opened 2023-12-11
Make sure disulphide search is converted to a SelectorBond.

### LINKED ISSUE #223: [BUG] Disulphide bond code fails when a single bond is present
The current logic for handling disulphide bonds will fail when a molecule only contains a _single_ bond. This is because the search returns a different object depending on the number of results that are found, i.e. a `SelectorBond` for multiple bonds, and a `Bond` for one bond. I need to check for the case of a single bond and convert the result into an iterable.

lohedges:
This PR closes #223 and closes OpenBioSim/biosimspace_tutorials#20 by making sure any (succesful) search for disulphide bonds is converted to a `SelectorBond`, thus is iterable. At present, if a single result is found, then it is returned as a `Bond`, hence the logic that follows the search will fail, i.e. getting an item from will extract an `Atom`, not a `Bond`.

I've tested the search logic for other bonds in a molecule, e.g. those between specific atom names, or indices. I don't (yet) have a single disulphide bond test molecule to add.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Parameters/_Protocol/_amber.py
+++ b/python/BioSimSpace/Parameters/_Protocol/_amber.py
@@ -725,7 +725,7 @@ class AmberProtein(_protocol.Protocol):
 
         # Try searching for disulphide bonds.
         try:
-            disulphides = query(mol, property_map)
+            disulphides = query(mol, property_map).bonds()
         except:
             disulphides = []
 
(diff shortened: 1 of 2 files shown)


lohedges:
Not sure if it's related to OpenMM 8.1, but those Exscientia restraint tests appear to be failing again (even with reduced tolerance).

chryswoods (approved):
All good - I accept this is confusing and a source of bugs. But, it makes for a much nicer interactive API. For scripts, there are the `.bond()` / `.bonds()` and associated functions that let us control what type of answer is needed from the query. I almost want a tool that could scan the code to auto-fix these bugs ;-)

PR #226, opened 2023-12-11
Fix issue #225

### LINKED ISSUE #225: [BUG] custom_parameters option not exposed via AMBER protein force field generator
The new `custom_parameters` keyword argument isn't exposed to AMBER protein force field generator `_make_amber_protein_function`, so is missing from the functions for these force fields.

lohedges:
This PR closes #225, by fixing the `custom_parameters` keyword argument for AMBER protein force fields. This option was added when decoupling from the additional `leap_commands` option, which now allows the user to reference the loaded molecule. Importantly, `custom_parameters` must be placed at the start of the LEaP script, i.e. before the molecule is loaded. The option should be passed as a list of paths to custom parameter files, which will be loaded with `loadAmberParams`, or the names of additional internal `leaprc` files, e.g. `leaprc.gaff`, which will be sourced with `source`. I've added a test that validates that both types of file can be handled and that the output in the `LEaP` script is in the correct order.

## Suggested reviewers:
@xiki-tempula

### DIFF
--- a/python/BioSimSpace/Parameters/_Protocol/_amber.py
+++ b/python/BioSimSpace/Parameters/_Protocol/_amber.py
@@ -227,14 +227,30 @@ class AmberProtein(_protocol.Protocol):
                     raise TypeError(
                         "'custom_parameters' must be a 'list' of 'str' types."
                     )
+                validated_parameters = []
                 for x in custom_parameters:
-                    if not os.path.isfile(x):
-                        raise ValueError(f"Custom parameter file does not exist: '{x}'")
+                    x.replace(" ", "")
+                    # Try to find built in force field files.
+                    if "leaprc" in x:
+                        ff = x.replace("leaprc.", "")
+                        try:
+                            ff = _find_force_field(ff)
+                            validated_parameters.append(f"source {ff}")
+                        except:
+                            raise ValueError(
+                                f"Custom parameter file does not exist: '{x}'"
+                            )
+                    else:
+                        if not _os.path.isfile(x):
+                            raise ValueError(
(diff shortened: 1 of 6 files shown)


xiki-tempula (approved):
LGTM

lohedges:
I've now split this into `pre_mol_commands` and `post_mol_commands`, which is what I was originally thinking of doing. The `pre_mol_commands` can be used for adding things like custom parameter files, whereas `post_mol_commands` can be used for anything that needs to reference the molecule itself.

This changes the API, so I should really do this as a feature. The issue is that the current 2023.4 implementation is broken, so I _do_ need to apply a fix there. Should I just revert the functionality there, i.e. move the position of the `leap_commands`? It would be a shame to need to apply this twice.

xiki-tempula:
Just revert to the old behaviour is good with me.

lohedges:
@chryswoods: Do you have any opinions on how to proceed? Should I merge this as a fix and backport to `main` as is, or merge as a feature and apply a separate fix to `main`. I don't think anyone is using this feature other than Exs and I assume that they are happy as long as it works. Given that the `custom_parameters` functionality is broken, I don't think anyone can be using it. (No reported issues.) If I were to apply a different fix on `main` then the easiest thing would be to remove the `custom_parameters` option entirely and revert the position at which the `leap_commands` are injected. (Annoying, my new solution is what I had originally proposed.)

chryswoods:
I think this can classify as a fix, so you can put it into the BioSimSpace patch release. It's at the borderline of being a feature, but if it isn't in widespread use, and we want a stable last patch while people play with 2023.5.0, then I think it is worth merging as a fix.

PR #237, opened 2024-02-05
Fix issue #236

### LINKED ISSUE #236: [BUG] GROMACS specific options are only applied for lambda=0 window when setup_only=True
As reported [here](https://github.com/OpenBioSim/biosimspace_tutorials/issues/23), process specific options such as `ingore_warnings` are only being used for the _first_ lambda window, so setup will fail for all other windows if warnings are present. We just need to apply these to the additional `grompp` commands used for the other windows.

lohedges:
This PR closes #236 and closes OpenBioSim/biosimspace_tutorials#23. This is achieved by making the `_generate_binary_run_file` method static, so that it can be called externally, e.g. during the `BioSimSpace.FreeEnergy.Relative` setup process. This is tested via the existing BioSimSpace tests, as well as the tutorial notebook for which the original issue was posted.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/FreeEnergy/_relative.py
+++ b/python/BioSimSpace/FreeEnergy/_relative.py
@@ -2060,9 +2060,7 @@ class Relative:
                 extra_lines=self._extra_lines,
                 property_map=self._property_map,
             )
-            if self._setup_only:
-                del first_process
-            else:
+            if not self._setup_only:
                 processes.append(first_process)
 
         # Loop over the rest of the lambda values.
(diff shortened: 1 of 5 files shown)


lohedges:
Now the Python 3.9 import test is repeatedly failing on Linux. (No error printed unfortunately.) Will need to debug locally to see what the issue is. Possibly some deprecated functionality in one of our dependencies?

lohedges:
The Windows issue appears to have magically fixed itself. The other failure is due to deprecated functionality in `pytest` 8 now rasing an exception. I'll need to pin to version 7 until `pytest-black` is updated. That said, it might be worth thinking about removing this from the build and using something like `pre-commit` instead.

lohedges:
Having fixed that, I now see the weird search result error on two builds :man_facepalming: 

```
FAILED tests/Sandpit/Exscientia/_SireWrappers/test_system.py::test_restraint_atoms[backbone-expected0] - assert [6, 8, 14, 15, 16] == [4, 5, 6, 8, 14, 15, 16]
  At index 0 diff: 6 != 4
  Right contains 2 more items, first extra item: 15
  Full diff:
  - [4, 5, 6, 8, 14, 15, 16]
  ?  ------
  + [6, 8, 14, 15, 16]
FAILED tests/_SireWrappers/test_system.py::test_restraint_atoms[backbone-expected0] - assert [6, 8, 14, 15, 16] == [4, 5, 6, 8, 14, 15, 16]
  At index 0 diff: 6 != 4
  Right contains 2 more items, first extra item: 15
  Full diff:
  - [4, 5, 6, 8, 14, 15, 16]
  ?  ------
  + [6, 8, 14, 15, 16]
```

lohedges:
And now everything passes :man_shrugging:

chryswoods:
Strange - I agree about switching to pre-commit. I moved to ruff locally, and its very small differences to black cause occasional reformats (and "reformat wars" on commonly edited files). A pre-commit would give us a canonical format that would be independent of person and IDE (i.e. independent of my local changes and inconsistencies ;-))

chryswoods (dismissed):
Looks good - I tend to use `Class.static_function()` when calling static functions, rather than `self.static_function()`, as it makes it more obvious that a static function is being called. On my quick scan I didn't get why we were passing lots of state from self into a self.function().

i.e. I would use `Gromacs._generate_binary_run_file(self._config_file...)` rather than `self._generate_binary_run_file(self._config_file...)`.

This is just my personal preference. I'm happy if you disagree and want to keep it as is.

lohedges:
No, that makes sense. In the cases where this is currently used the method is still called on an instance, but a class method makes more sense for generality (and to make testing easier).

lohedges:
In this use case I create an initial process that is the reference for the others, hence re-use all of its attributes when re-running grompp, e.g. making sure we use the same executable, etc.

lohedges:
In this use case I create an initial process that is the reference for the others, hence re-use all of its attributes when re-running grompp, e.g. making sure we use the same executable, etc. As the instance exists, I just called the method on it. I agree it would be clearer to just call on the class directly.

PR #244, opened 2024-02-19
Fix issues #241, #242, and #243

### LINKED ISSUE #241: [BUG] readPerturbableSystem needs to remove shared system properties
Like `BioSimspace.IO.readMolecules`, the `BioSimSpace.IO.readPerturbableSystem` function needs to remove any shared properties that are automatically added by Sire. These cause issues when combining molecules into other systems.

### LINKED ISSUE #242: [BUG] OpenFF parameterisation error reports the incorrect file name.
Since switching over to using `openff-interchange`, we have forgotten to update the file name reported in an OpenFF parameterisation error. (Previously `parmed` was used, so the files had a `parmed` prefix.)

### LINKED ISSUE #243: [BUG] Expose FreeEnergyMixin protocols for GROMACS
While the `FreeEnergyMixin` protocols have been added, e.g. `FreeEnergyMinimisation`, these haven't yet been exposed for engines that support them. (Currently only GROMACS.)

lohedges:
This PR...

closes #241
    This removes shared properties when loading a perturbable system from file. This is needed to avoid issues when combining molecules together to form new systems. I've also added some extra logic into the GROMACS process code to check for a periodic system when updating the space property. For vacuum simulations we use a "fake" infinite box (as recommended by GROMACS) and I don't want to copy this property back if the original system contained an infinite cartesian space, which is now the default in Sire. Note that there was a typo in the existing code for `readMolecules`, which didn't cause any issues in practice since the property isn't used internally.

closes #242
    This updates the error message to use the correct file names and removes the now redundant `parmed` import.

closes #243
    This exposes the `FreeEnergyMixin` protocols, allowing GROMACS to run any existing protocol and intermediate lambda values, not just the end states. (This will be also be exposed for AMBER when ready.)

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/IO/_io.py
+++ b/python/BioSimSpace/IO/_io.py
@@ -566,7 +566,7 @@ def readMolecules(
         prop = property_map.get("time", "time")
         time = system.property(prop)
         system.removeSharedProperty(prop)
-        system.setProperties(prop, time)
+        system.setProperty(prop, time)
     except:
         pass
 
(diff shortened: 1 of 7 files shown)


chryswoods (approved):
Looks good. I think this is good motivation to (in future) add a "unshareProperties" or similar function to the sire Molecule so that it can automatically unshare all of its properties.

PR #257, opened 2024-03-08
Only convert to end state if this isn't a FreeEnergy protocol.

### LINKED ISSUE #256: [BUG] GROMACS hybrid topology file is lost when protocol used position restraints
When adding position restraints we need to write a modified GROMACS topology file. Incorrectly, we always convert to the lambda = 0 state prior to writing the topology, rather than checking if this is  `FreeEnergyProtocol` first, in which case the hybrid topology should be written.

lohedges:
This PR closes #256 by checking for a `FreeEnergy` protocol before extracting an end state prior to writing the modified topology file when position restraints are used.

## Suggested reviewers:
@chryswoods


chryswoods (approved):
👍

PR #260, opened 2024-03-08
Fix issues #258 and #259

### LINKED ISSUE #258: [BUG] Ions should be excluded in AMBER position restraint mask
Currently we explicitly list atom indices when under the restraint mask size limit, but use a name matching mask when exceeding this. Here the mask should exclude ions along with water molecules. For simplicity (since we need to match by name) we should just exclude Na and Cl, since this means that a BioSimSpace workflow will be okay.

### LINKED ISSUE #259: [BUG] The __pow__ operator for BioSimSpace.Types._GeneralUnit doesn't handle fractional exponents
Currently the `__pow__` operator for `BioSimSpace.Types._GeneralUnit` doesn't support fractional exponents, e.g. `unit ** (1/2)`

lohedges:
This PR...

1) closes #258 by excluding the standard free ions (Na and Cl) to the AMBER position restraint mask. This logic is triggered when the manual restraint mask exceeds the mask character limit, so we need to fall back on the assumption of _standard_ AMBER atom naming. This will work in the case of a BioSImSpace workflow, e.g. where solvation is performed with BioSimSpace, but not for arbitrary input. (The restraint is unreliable then anyway.)

2) closes #259 by adding support for fractional exponents in the `BioSimSpace.Types._GeneralUnit.__pow__` operator. I've also re-orded the dimension mask used internally for consistency with sire, which will make it easier to create `GeneralUnit` types from these masks. (I hadn't used this before.) These update is largely covered by the existing tests but I've added a further test specific to fractional exponents to cover a range of cases, including error handling.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Process/_gromacs.py
+++ b/python/BioSimSpace/Process/_gromacs.py
@@ -1995,8 +1995,10 @@ class Gromacs(_process.Process):
             # Create a copy of the system.
             system = self._system.copy()
 
-            # Convert to the lambda = 0 state if this is a perturbable system.
-            system = self._checkPerturbable(system)
+            # Convert to the lambda = 0 state if this is a perturbable system and this
+            # isn't a free energy protocol.
+            if not isinstance(self._protocol, _FreeEnergyMixin):
+                system = self._checkPerturbable(system)
 
             # Convert the water model topology so that it matches the GROMACS naming convention.
             system._set_water_topology("GROMACS")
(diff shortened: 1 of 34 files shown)


lohedges:
Looks like I just need to update a unit mask in the ABFE restraint code in the Sandpit. Will fix on Monday.

lohedges:
Strangely it looks like `requests` is missing on Windows 3.11.

chryswoods (dismissed):
Looks great - I like how pow has let you remove special case code elsewhere. Nice use of rounding. I need to remember to port this down into sire too.

Good unit tests :-)

lohedges:
Thanks. Yes, this simplified a lot of stuff elsewhere so it was worth doing, regardless of whether it will be needed in the long term.

PR #269, opened 2024-04-05
Make handling of base units more robust

fjclark:
This ensures that only classes, and not modules, end up in the _base_units list. I am trying to integrate BSS-based code with [Maize](https://github.com/MolecularAI/maize/tree/public), and (during testing only) the _base_units_list would end up containing modules rather than just classes. Maize does a lot of messing around with the environment (it failed [here](https://github.com/MolecularAI/maize/blob/490ea9481e2afbb2336a683474cfe393c69f4b25/maize/core/node.py#L518)).

Also, Maize looks like a really neat solution for running BSS nodes/ workflows, with options to submit through slurm etc. I'm planning to convert some BSS nodes to Maize nodes and implement some workflows over the next few weeks.

## Suggested reviewers:
@lohedges, @chryswoods

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/Types/_base_units.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/Types/_base_units.py
@@ -24,6 +24,8 @@
 
 __all__ = ["_base_units", "_base_dimensions", "_sire_units_locals"]
 
+import sys as _sys
+
 from ._angle import *
 from ._area import *
 from ._charge import *
@@ -34,12 +36,12 @@ from ._temperature import *
 from ._time import *
 from ._volume import *
 
-import sys as _sys
-
 _namespace = _sys.modules[__name__]
 
 # Create the list of base unit types.
-_base_units = [getattr(_namespace, var) for var in dir() if var[0] != "_"]
+_base_units = [
+    getattr(_namespace, var) for var in dir() if var[0] != "_" and isinstance(var, type)
+]
 
 _base_dimensions = {}
 for unit in _base_units:
--- a/python/BioSimSpace/Sandpit/Exscientia/Types/_base_units.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/Types/_base_units.py
@@ -38,10 +38,9 @@ from ._volume import *
 
 _namespace = _sys.modules[__name__]
 
-# Create the list of base unit types.
-_base_units = [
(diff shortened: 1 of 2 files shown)


fjclark:
Sorry - forgot to test locally before the initial commit.

lohedges:
Thanks, I'll take a look on Monday. Maize looks interesting. Hopefully it's more useful than my CWL interface, which I think no-one has used ;-) Let me know if you end up needing some interface to set cluster specific settings, e.g. partitions, hardware, etc. A while back I was thinking about wrapping `dask.Jobqueue` and wanted a way for a user (or sys admin) to be able to configure settings that would be used by nodes, etc.

lohedges:
Actually, this won't work if the object isn't a type, so we'd also need to filter for that first, i.e. a double filter.

lohedges:
```
diff --git a/python/BioSimSpace/Types/_base_units.py b/python/BioSimSpace/Types/_base_units.py
index 931668f8..267cb551 100644
--- a/python/BioSimSpace/Types/_base_units.py
+++ b/python/BioSimSpace/Types/_base_units.py
@@ -24,6 +24,8 @@

 __all__ = ["_base_units", "_base_dimensions", "_sire_units_locals"]

+import sys as _sys
+
 from ._angle import *
 from ._area import *
 from ._charge import *
@@ -33,13 +35,17 @@ from ._pressure import *
 from ._temperature import *
 from ._time import *
 from ._volume import *
-
-import sys as _sys
+from ._type import Type as _Type

 _namespace = _sys.modules[__name__]

 # Create the list of base unit types.
 _base_units = [getattr(_namespace, var) for var in dir() if var[0] != "_"]
+# Filter out any non-Type objects. (This can happen when BioSimSpace is
+# wrapped by other tools, e.g. Maize.)
+_base_units = [
+    unit for unit in _base_units if isinstance(unit, type) and unit.__base__ == _Type
+]
```

fjclark:
> Do you know if Maize will only ever import things as modules, or could it import objects directly?

I'm not sure - haven't had enough of a detailed look at the code yet. This looks great though, and works for me - thanks for the quick fix!

fjclark:
>Let me know if you end up needing some interface to set cluster specific settings, e.g. partitions, hardware, etc

Thanks. I think this all should be handled through Maize, which uses https://github.com/ExaWorks/psij-python for job submission.

PR #271, opened 2024-04-09
Fix issue #270

### LINKED ISSUE #270: [BUG] Non-reproducible search for restraint atoms
Searching for restraint atoms using `system.getRestraintAtoms("backbone")` gives non-reproducible results due to the use of a set to hold the protein and nucleic acid residue names. This has been causing intermittent test failures in the CI for some time. Switching to a list for the container fixes things and the tests reproducibly pass.

One concern is that the search result appears to be _order_ dependent. I'm not yet sure if this is a bug, or whether Sire has an upper limit on things `resname` searches where the items are comma separated. If so, then we can't guarantee the result. (Note that the examples in the tests _do_ pass.)

lohedges:
This PR closes #270 by using a list rather than a set to store the names of protein and nucleic acid residues. This makes the search string for backbone atoms reproducible, hence the tests now repeatedly pass. I'm still not sure why this issue has only cropped up recently, since `_prot_res` has been a set since the code was first added.

```python
In [1]: x = {"1", "2", "3"}

In [2]: ",".join(x)
Out[2]: '2,3,1'

In [3]: x = ["1", "2", "3"]

In [4]: ",".join(x)
Out[4]: '1,2,3'
```

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_molecule.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_molecule.py
@@ -792,7 +792,11 @@ class Molecule(_SireWrapper):
             if len(matches) < num_atoms0:
                 # Atom names or order might have changed. Try to match by coordinates.
                 matcher = _SireMol.AtomCoordMatcher()
-                matches = matcher.match(mol0, mol1)
+
+                try:
+                    matches = matcher.match(mol0, mol1)
+                except:
+                    matches = []
 
                 # We need to rename the atoms.
                 is_renamed = True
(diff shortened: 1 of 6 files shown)


lohedges:
Actually, it appears that the issue was that the protein and nucleic acid residue strings weren't being joined correctly. No comma was added between then so the _last_ protein residue name was being combined with the _first_ nucleic acid residue. Depending on the ordering from the set, this could result in removing some of the matching residues from the test molecule, e.g. `ACE` combined with `T` would create a nonsense residue called `ACET`, so we would no longer match `ACE`.

chryswoods (approved):
All good - the change from `{ }` to `[ ]` is very subtle - good spot that this was one of the causes of the inconsistent behaviour. Same with the missing "," in the join.

PR #276, opened 2024-04-12
Final fixes for the 2024.1.0 release

### LINKED ISSUE #273: [BUG] fileformat property no longer preserved when converting a molecule to a system
This used to be the case, but no longer works for some reason. I think I must have removed the manual property setting when the new `shared_property` code was added. This preserves the property when a molecule is removed from the system, but not when that molecule is converted back into a system. This is an easy fix. (We only do this for single molecule systems.)

### LINKED ISSUE #274: [BUG] The base Protocol should return None for getRestraint
Currently the `getRestraint` method is only present for protocols that are instances of `_PositionRestraintMixin`. This means that some of the config generators fail when trying to call `getRestraint` on a pure protocol class. We should add a method to the base class that returns `None`, unless the object is an instance of `_PositionRestraintMixin`.

### LINKED ISSUE #275: [BUG] Protocol.Metadynamics is missing the thermostat_time_constant option
As in the description.

lohedges:
This PR adds some final fixes ahead of the 2024.1.0 release. Once complete, I'll synchronise with the Exscientia remote and do the release PR.

Changes are:

* Switch to using `os.path.join` to create file names throughout. I haven't updated in the Sandpit too, since they only use macOS and Linux so are unaffected. I don't think the existing code was a problem, but this should hopefully avoid any future issues. (Most of the functionality that was updated doesn't work on Windows anyway.)
* Closes #273 by adding the `fileformat` property to systems that are created from a single molecule. (This caused the metadynamics demo to fail, since the file format of an extracted molecule was used to determine the best engine to run the simulation with.)
* Closes #274 by making sure that _all_ protocols have a `getRestraint` method, which returns `None` if it isn't derived from `PositionRestraintMixin`.
* Closes #275 by adding the missing `thermostat_time_constant` option to the metadynamics protocol.
* I've refactored the test fixtures to move the common systems into the `conftest` file.
* I've added tests for steered molecular dynamics, metadynamics, and funnel metadynamics. The first two are only run locally, since they require a PLUMED patched version of GROMACS. These tests will make it easier to catch issues with the tutorials suite.

Checklist:

## Suggested reviewers:
@chryswoods

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,35 +63,35 @@ Conda package
 The easiest way to install BioSimSpace is using our `conda channel <https://anaconda.org/openbiosim/repo>`__.
 BioSimSpace is built using dependencies from `conda-forge <https://conda-forge.org/>`__,
 so please ensure that the channel takes strict priority. We recommend using
-`Mambaforge <https://github.com/conda-forge/miniforge#mambaforge>`__.
+`Miniforge <https://github.com/conda-forge/miniforge>`__.
 
 To create a new environment:
 
 .. code-block:: bash
 
-    mamba create -n openbiosim -c conda-forge -c openbiosim biosimspace
-    mamba activate openbiosim
+    conda create -n openbiosim -c conda-forge -c openbiosim biosimspace
+    conda activate openbiosim
 
 To install the latest development version you can use:
 
 .. code-block:: bash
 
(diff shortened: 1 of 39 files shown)


chryswoods:
I've installed and am just running the tests locally now. It's all looking good :-)

PR #277, opened 2024-04-15
Sync with Exscientia remote

lohedges:
This PR synchronises with the [Exscientia remote](https://github.com/Exscientia/BioSimSpace) ahead of the 2024.1.0 release. This builds on top of #276. This has been tested locally but the CI will determine if there are any unguarded tests or imports in the synchronised code. (I think I've caught them all, but there always seems to be something.)


lohedges:
Nope, missed one. Really wish this could be automated.

PR #287, opened 2024-05-07
Fix issue #284

### LINKED ISSUE #284: Option to write parameterisation intermediary file in SDF format to retain bond information
Hey @lohedges,

I had an issue with a molecule I was trying to parameterise using gaff2. The antechamber error that arose was:

> This molecule may have more than one unit.
>        antechamber can only handle one unit.  If the input is a single unit
>        then the connectivity is wrong and the geometry may be bad.
>        Please convert your molecule to a mol2 file via:
>        antechamber -j 5 -at sybyl -dr no 
>        And then check your molecule with a visualization program;
>        manually add missing bonds or delete unwanted bonds as appropriate.

I knew this error was slightly non-sensical because the input file is an SDF (that can be fully sanitized using rdkit, and so the bond info is definitely correct). Looking at the source code I saw that BSS [writes the molecule to PDB](https://github.com/OpenBioSim/biosimspace/blob/9b605e8b33e2d900951e879d9cb105286313c8b5/python/BioSimSpace/Parameters/_Protocol/_amber.py#L1086C1-L1108C1) before parsing this file to antechamber.

```py
        # Write the system to a PDB file.
        try:
            _IO.saveMolecules(
                _os.path.join(str(work_dir), "antechamber"),
                new_mol,
                "pdb",
                property_map=self._property_map,
            )
        except Exception as e:
            msg = "Failed to write system to 'PDB' format."
            if _isVerbose():
                msg += ": " + getattr(e, "message", repr(e))
                raise IOError(msg) from e
            else:
                raise IOError(msg) from None

        # Generate the Antechamber command.
        command = (
            "%s -at %d -i antechamber.pdb -fi pdb "
            + "-o antechamber.mol2 -fo mol2 -c %s -s 2 -nc %d"
        ) % (_antechamber_exe, self._version, self._charge_method.lower(), charge)

```

Since the PDB doesn't contain bonding information, antechamber appears to be unable to correctly interpret the atom distances to approximate the bonds. Since the bond is present in the molecule (mol._sire_object.property("connectivity")) I figured it should be possible to write an intermediary SDF and use this as input to antechamber. (On another note I see that the 'bond' property isn't available for this SDF loaded molecule, even though 'connectivity' is. Would you be able to explain the logic here?)

I therefore monkey-patched the GAFF class with a new run method, simply replacing the PDB format with SDF format:
 ```py
# Write the system to a PDB file.
        try:
            _IO.saveMolecules(
                _os.path.join(str(work_dir), "antechamber"),
                new_mol,
                "sdf",
                property_map=self._property_map,
            )
        except Exception as e:
            msg = "Failed to write system to 'PDB' format."
            if _isVerbose():
                msg += ": " + getattr(e, "message", repr(e))
                raise IOError(msg) from e
            else:
                raise IOError(msg) from None

        # Generate the Antechamber command.
        command = (
            "%s -at %d -i antechamber.sdf -fi sdf "
            + "-o antechamber.mol2 -fo mol2 -c %s -s 2 -nc %d"
        ) % (_antechamber_exe, self._version, self._charge_method.lower(), charge)
```

This appeared to work nicely and resolve the antechamber issue.

I therefore wonder whether it would be possible to check if the input molecule has connectivity / bond information, and if it does write an intermediary SDF instead of PDB file. I understand that SDFs are unique in that they're parsed to BSS as both coordinate and topology files and so this might just be appropriate for my use case, but I feel like SDF is becoming the common way of storing molecules. Unfortunately the molecule I'm using that throws the antechamber error is proprietary so I can't share the structure.

Would be good to hear your thoughts!

Cheers,
Noah

lohedges:
This PR closes #284 by using SDF as the intermediate format for `antechamber` if the original molecule was loaded from that format, otherwise PDB is used. This mirrors the functionality used for the OpenFF protocol. A more general solution could be to allow the user to specify the intermediate format, but I'm not sure if this is needed at this time.

The fix has been tested by the author of issue #284. They are unable to share the molecule in question so I haven't added a unit test at this time.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Parameters/_Protocol/_amber.py
+++ b/python/BioSimSpace/Parameters/_Protocol/_amber.py
@@ -1083,12 +1083,23 @@ class GAFF(_protocol.Protocol):
                     )
                     raise _ParameterisationError(msg)
 
-        # Write the system to a PDB file.
+        # If the molecule was loaded from an SDF file, then use this as the
+        # intermediate format.
+        fileformat_prop = self._property_map.get("fileformat", "fileformat")
+        if (
+            new_mol._sire_object.hasProperty(fileformat_prop)
+            and "SDF" in new_mol._sire_object.property("fileformat").value()
+        ):
+            format = "sdf"
+        else:
+            format = "pdb"
+
+        # Write the system to file.
         try:
             _IO.saveMolecules(
                 _os.path.join(str(work_dir), "antechamber"),
                 new_mol,
-                "pdb",
+                format,
                 property_map=self._property_map,
             )
         except Exception as e:
@@ -1101,9 +1112,16 @@ class GAFF(_protocol.Protocol):
 
         # Generate the Antechamber command.
         command = (
-            "%s -at %d -i antechamber.pdb -fi pdb "
+            "%s -at %d -i antechamber.%s -fi %s "
             + "-o antechamber.mol2 -fo mol2 -c %s -s 2 -nc %d"
-        ) % (_antechamber_exe, self._version, self._charge_method.lower(), charge)
+        ) % (
+            _antechamber_exe,
+            self._version,
+            format,
+            format,
+            self._charge_method.lower(),
+            charge,
+        )
 
         with open(_os.path.join(str(work_dir), "README.txt"), "w") as file:
             # Write the command to file.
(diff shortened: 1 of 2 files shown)


chryswoods (approved):
👍

PR #289, opened 2024-05-13
Fix issue #288

### LINKED ISSUE #288: [BUG] Dummy atoms not detected during pert file writing when system created via sire.morph.create_from_pertfile
When a pert file is written by BioSimSpace, bonded terms involving dummy atoms are written to the file meaning that these terms will be constrained. Dummy atoms are detected when an atom has a dummy element for either its `element0` or `element1` property. Following the writing of the pert file, the element property of these atoms is modified to take the value of the end state with the most protons. This means that the dummy element property is lost when a perturbable molecule is reconstructed using `sire.morph.create_from_pertfile`, and subsequently the dummy bonded terms won't be written if that molecule is re-used as input to a `somd` simulation in BioSimSpace, i..e. the pert file isn't round-trip self-consistent.

A simple solution is to also check for dummy atoms using the `ambertype` property, which will remain `du` for dummy atoms when reconstructed via sire.

lohedges:
This PR closes #288 by checking both the element _and_ ambertype properties when detecting dummy atoms. This avoids a round-trip self-consistency issue when creating `somd1` input from perturbable molecules created via `sire.morph.create_from_pertfile`, where the element property will have been replaced with the value of the end state with the largest number of protons, i.e. dummies may no longer have `Element("Xx")`.

This PR supersedes #287.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Parameters/_Protocol/_amber.py
+++ b/python/BioSimSpace/Parameters/_Protocol/_amber.py
@@ -1083,12 +1083,23 @@ class GAFF(_protocol.Protocol):
                     )
                     raise _ParameterisationError(msg)
 
-        # Write the system to a PDB file.
+        # If the molecule was loaded from an SDF file, then use this as the
+        # intermediate format.
+        fileformat_prop = self._property_map.get("fileformat", "fileformat")
+        if (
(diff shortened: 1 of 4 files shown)


chryswoods (approved):
Very subtle bug and good fix :-)

PR #294, opened 2024-05-21
Add Protein FEP Functionality

akalpokas:
This PR introduces protein FEP functionality to BioSimSpace and allows for creating of hybrid protein/peptide systems with single or multiple simultaneous modifications, which can include point mutations to canonical or non-canonical amino acids, as well as covalent modifications.

The PR is essentially based around region of interest (ROI) idea, and code wise does the following:

- Moves the `merge` function from [Exscientia sandpit](https://github.com/OpenBioSim/biosimspace/blob/devel/python/BioSimSpace/Sandpit/Exscientia/Align/_merge.py) out of the sandpit into the standard merge code location. I have tested this code with ethane to methanol perturbation, and it gives the same exact perturbation as the non-sandpit merge code, (I looked at gromacs .top files specifically) although it is probably a good idea to test this with more complex ligand perturbations.
- This merge code is then modified to remove the need to provide ROI atom indices as inputs (this was a requirement in original Exscientia's implementation). Instead these are reconstructed inside the merge function which simplifies the overall workflow for the user, as now they only need to pass the ROI residue indices. The merge function is then modified to allow for multiple simultaneous merges to be performed between two molecules. This was tested with MDM2 T16G-I19G double mutant system and behaves as intended. Note that these modifications are only invoked when using ROI code only, it does not affect the standard merge code.
- The `roiMatch` function is then added to the Align module. This function rapidly computes mappings between two large input molecules based on the idea of region of interest. It supports standard BioSimSpace.Align.matchAtoms as well as [Kartograf](https://github.com/OpenFreeEnergy/kartograf) as mapping backends. Therefore an internal `_kartograf_map` function is also added to the Align module. In my experience kartograf in the protein FEP context is really only needed for niche cases such when trying to compute a mapping between two enantiomers of two covalently modified residues, so it's not critical to have it as a mapping backend for most use cases. The `matchAtoms` function is **also modified in order to expose some of more lower-level rdKit mapping functionality** which I found useful sometimes when trying to force a specific mapping for covalently modified residues. It does not affect the default behaviour of the mapping function.
- The `roiAlign` function is also added to the Align module. This function aligns selected residues from molecule0 to molecule1, which allows us to drop the assumption that two input proteins need to be in the same conformation in order to create a merged protein. In most cases this assumption is satisfied (if you just create a carbon copy of the wild-type protein and mutate 1 or more residues), however with this function we can execute more complex workflows such as non-equilibrium switching protein FEP (where you would generate wild-type and mutant trajectories separately without alchemical code, and then start to create hybrid snapshots that now have to deal with two input protein structures which are not fully aligned). This function can either use `rmsdAlign` or `flexAlign` functions internally, depending on how precise alignment is needed. In my experience `rmsdAlign` works perfectly after some minimisation.

Overall, this means that a hybrid proteins can be created with just few lines of code that are nearly identical to the normal input creation for FEP workflow:

```python
import BioSimSpace as BSS
mdm2_wt = BSS.IO.readMolecules(f"mdm2_wt.pdb")[0]
mdm2_mut = BSS.IO.readMolecules(f"mdm2_v14g.pdb")[0]

mdm2_wt = BSS.Parameters.ff14SB(mdm2_wt, ensure_compatible=False).getMolecule()
mdm2_mut = BSS.Parameters.ff14SB(mdm2_mut, ensure_compatible=False).getMolecule()

mapping = BSS.Align.roiMatch(molecule0=mdm2_wt, molecule1=mdm2_mut, roi=[9])
aligned_wt = BSS.Align.roiAlign(molecule0=mdm2_wt, molecule1=mdm2_mut, roi=[9])
merged = BSS.Align.merge(molecule0=aligned_wt, molecule1=mdm2_mut, mapping=mapping, roi=[9])
```
A minimal code example with input files is also provided:  [minimal_pfep_example.tar.gz](https://github.com/OpenBioSim/biosimspace/files/15392209/minimal_pfep_example.tar.gz)

While in theory the `roiMatch` and `roiAlign` functions could be incorporated to already existing `matchAtoms` and `rmsdAlign` functions to have ROI functionality (in the way that `merge` function has ROI functionality and not a separate `roiMerge` function), I opted to have them as separate functions essentially to eliminate any risk of inadvertently modifying their behaviour and introducing bugs down the line. I think that this approach is safer overall, even if it introduces more functions to the Align module.

Another thing to note is that `viewMapping` function will be quite useless with this current ROI implementation since the mappings provided to the function will usually be quite large, in the future I would like to modify this function to allow mapping viewing between ROI residues. It's also important to note that I haven't tested this implementation with every possible canonical amino acid mutation, therefore it should be still treated as somewhat experimental, although I do believe the overall workflow will work correctly and any potential issues would arise from how the residues are mapped with the backend MCS function. Proline mutations are of course still not supported.
___

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): :white_check_mark:
* I confirm that I have added a test for any new functionality in this pull request: :x: - I still need to write automated tests for the new code, although the code does contain some sanity checks to test whether same residues are being mapped, etc.
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: :x: - This still needs to be done too. 
* I confirm that I have permission to release this code under the GPL3 license: :white_check_mark:

## Suggested reviewers:
@lohedges,


akalpokas:
TODO before merging:

- [ ] Write tests for the newly implemented ROI functions
- [ ] Consolidate ROI functions into standard `BSS.Align` functions (i.e. `roiMatch(roi=x)` --> `matchAtoms(roi=x)`
- [ ] Update `BSS.Align.viewMapping` to support viewing ROIs

lohedges:
Suggested refactoring if we wish to preserve separate matching and alignment functions for regular and roi implementions...

First move the existing functions to private module only implementations (not exposed to the user), e.g.:

```python
def matchAtoms(...): --> def _match_atoms(...):

def roiMatch(...): --> def _match_roi(...):

def rmsdAlign(...): --> def _rmsd_align(...):

def roi_align(...): --> def _roi_align(...):
```

Then you just need to create wrapper functions that match the original APIs of `matchAtoms` and `rmsdAlign` with the addition of the extra `roi` kwarg. Inside these functions we check if `roi` is None, then call the appropriate backend, i.e.:

```python
# Same as before, but with extra roi kwarg.
def matchAtoms(..., roi=None):
    """Same docstring as before, plus roi kwarg."""
   
    # Use regular backend.
    if roi is None:
        return _match_atoms(...)
    # Use ROI backend.
    else:
        return _match_roi(..., roi=roi)

# Same thing for rsmdAlign.
def rmsdAlign(..., roi=None):
    """Same docstring as before, plus roi kwarg."""
   
    # Use regular backend.
    if roi is None:
        return _rmsd_align(...)
    # Use ROI backend.
    else:
        return _roi_align(..., roi=roi)
```

akalpokas:
Thanks for your suggestions @lohedges. I have now refactored the code to work as one set of functions, and added tests:

```python
import BioSimSpace as BSS
mdm2_wt = BSS.IO.readMolecules(f"mdm2_wt.pdb")[0]
mdm2_mut = BSS.IO.readMolecules(f"mdm2_v14g.pdb")[0]

mdm2_wt = BSS.Parameters.ff14SB(mdm2_wt, ensure_compatible=False).getMolecule()
mdm2_mut = BSS.Parameters.ff14SB(mdm2_mut, ensure_compatible=False).getMolecule()

mapping = BSS.Align.matchAtoms(molecule0=mdm2_wt, molecule1=mdm2_mut, roi=[9])
aligned_wt = BSS.Align.rmsdAlign(molecule0=mdm2_wt, molecule1=mdm2_mut, roi=[9])
merged = BSS.Align.merge(molecule0=aligned_wt, molecule1=mdm2_mut, mapping=mapping, roi=[9])
```

In addition `BSS.Align.viewMapping` can now be used to highlight specific residues of interest:

```python
BSS.Align.viewMapping(mdm2_wt, mdm2_mut, mapping) # without a specified residue
```
image

```python
BSS.Align.viewMapping(mdm2_wt, mdm2_mut, mapping, roi=0) # highlighting a wild-type residue
```
image

```python
BSS.Align.viewMapping(mdm2_wt, mdm2_mut, mapping, roi=9) # highlighting a mutant residue
```
image

I was also hoping to display the rest of the protein either in cartoon style or in stick style with reduced opacity, however the former can't be done because  py3Dmol sees the input structure as a "small molecule" rather than a protein (which I think happens due to the conversion to rdKit molecule in the function). The latter doesn't seem to work because py3Dmol doesn't allow mixing opacities for a single style. I'm happy to make changes to the way the function displays the mapping if needed.

python/BioSimSpace/Align/_align.py
```diff
@@ -1297,13 +1763,140 @@ def flexAlign(
...
+
+    if not isinstance(molecule1, _Molecule):
+        raise TypeError(
+            "'molecule1' must be of type 'BioSimSpace._SireWrappers.Molecule'"
+        )
+    if roi is None:
+        raise ValueError("No region of interest (ROI) has been provided.")
+    else:
+        if not isinstance(roi, list):
+            raise TypeError("'roi' must be of type 'list'.")
```
lohedges, line 1836: Extra validation of `roi` as described in previous comment.

python/BioSimSpace/Align/_merge.py
```diff
@@ -74,6 +75,10 @@ def merge(
         takes precedence over 'allow_ring_breaking' and
         'allow_ring_size_change'.
 
+    roi : list
+        The region of interest to merge.
+        Consists of a list of ROI residue indices.
+
```
lohedges, line 81: The `roi` kwarg doesn't appear to be validated at the start of the function.

lohedges (changes requested):
This is looking really nice. I've added a bunch of minor comments. These should be easy to address. Let me know if you still plan on updating the visualisation code. I'm happy to take a look at this too.

I'll take another read through the code later. I've added your remote locally so I'll also play around with the tests too, e.g. seeing what happens when I give nonsensical input, etc.

python/BioSimSpace/Align/_align.py
```diff
@@ -1297,13 +1763,140 @@ def flexAlign(
...
+            else:
+                mol0_coords.append(res.property("coordinates"))
+
+        # Flatten the list
+        mol0_coords = [item for sublist in mol0_coords for item in sublist]
+
+        # Create a cursor for updating the coordinates property
+        c = mol0.cursor()
+        for i, atom in enumerate(c.atoms()):
+            atom["coordinates"] = mol0_coords[i]
```
lohedges, line 1882: The property might not be called `coordinates`, so we need to use the passed map to get the actual name, i.e. `property_map0.get("coordinates", "coordinates")`.

python/BioSimSpace/Align/_align.py
```diff
@@ -1030,7 +1080,345 @@ def matchAtoms(
...
+
+    return kartograf_mapping
+
+
+def _roiMatch(
+    molecule0,
+    molecule1,
+    roi,
+    use_kartograf,
+    kartograf_kwargs,
```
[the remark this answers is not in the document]
akalpokas: Thanks, I changed these to keyword arguments

python/BioSimSpace/Align/_align.py
```diff
@@ -1030,7 +1080,345 @@ def matchAtoms(
...
+            "'molecule0' must be of type 'BioSimSpace._SireWrappers.Molecule'"
+        )
+
+    if not isinstance(molecule1, _Molecule):
+        raise TypeError(
+            "'molecule1' must be of type 'BioSimSpace._SireWrappers.Molecule'"
+        )
+
+    if roi is type(list):
+        raise TypeError("'roi' must be of type 'list'")
```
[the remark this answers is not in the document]
akalpokas: I added `_validate_roi()` function in the `Align` module that checks for this.

python/BioSimSpace/Align/_align.py
akalpokas: Same as in the previous comment.

python/BioSimSpace/Align/_merge.py
akalpokas: I added validation code to check for this, however I cannot import `_validate_roi()` from the `Align` module because `Align` module itself imports the `Merge` module, which would result in a circular import. So at the moment it's just the code duplicated from the `_validate_roi()` function. I could move this function to the `Merge` module, but the `Align` module has `_validate` functions in it already, so I thought it would be better left in there.

python/BioSimSpace/Align/_align.py
akalpokas: Thanks, I added code to get the property name from the map if provided.

python/BioSimSpace/Align/_merge.py
lohedges: You should just be able to do an in-pace import, i.e. don't import `_validate_roi` at the top of the module, rather just before it's used. That way it's not imported until the point at which it's used, rather than during the initial module import. (I need to refactor the code to do this for everything, since it will greatly reduce the initial import time.)

akalpokas: I tried doing that before, basically by having something like this in the `merge` function:
```python
# Validate the region of interest.
if roi is not None:
    from ..Align import _validate_roi
    _validate_roi(roi, molecule0, molecule1)
```
But this gives me:
```python
ImportError: cannot import name '_validate_roi' from 'BioSimSpace.Align' 
(/home/akalpokas/code/git/biosimspace_fork/python/BioSimSpace/Align/__init__.py)
```

I think this could be solved if I had `_validate_roi` inside the `__init__.py`, but it doesn't seem like a good place for a private function.

akalpokas:
As per our discussions, I added the code that replaces old `Align.viewMapping()` code with the [code used by OFE](https://github.com/OpenFreeEnergy/gufe/blob/12bc6447c9886d7fdb8fbdb6e62ee061ac0207d8/gufe/visualization/mapping_visualization.py). This simplifies the viewing of the mapping and allows us to view it in 2D instead of 3D. The code is still very much WIP and needs a lot of refactoring, but basically viewing the mapping for ethane to methanol perturbation looks like this:

```python
BSS.Align.viewMapping(ethane, methanol, mapping) # default image size is 300 pixels
```
image

```python
BSS.Align.viewMapping(ethane, methanol, mapping, pixels=400) # increase image size
```
image

For viewing residues in the protein:
```python
BSS.Align.viewMapping(ala_mut, ala_wt, mapping, roi=3, pixels=300)
```
image

```python
BSS.Align.viewMapping(ala_mut, ala_wt, mapping, roi=3, pixels=500, show_adjacent_residues=True) # show neighbouring residues
```
image

python/BioSimSpace/Align/_merge.py
lohedges: I think you just want to import directly from the module, i.e.:

```
from ._align import _validate_roi
```

lohedges:
Brilliant, those look really snazzy.

lohedges:
I've just merged some fixes into `devel`so you'll need to merge across to your branch and resolve any conflicts. The only one that could cause an issue is [this](https://github.com/OpenBioSim/biosimspace/pull/295). If conflicts do arise, you should be able to look at the block of code added in the PR and just paste it at the right place in your edited file. Let me know if there are any problems.

lohedges:
I've triggered the CI and all tests are passing, other than an unreleated IO error on Windows which I'm re-running now. (This is a periodic problem we see with the WIndows runners and is nothing to do with your code.)

Just a couple of things to check before merging:

1) Did you manage to fix the circular import issue that you mentioned above, i.e. importing `_validate_roi` directly from the module file.
2) Are you happy with the updated visualisation code, or does it require more polish before merging? As long as it works things should be okay. It can always be refactored at a later date. (Assuming the API doesn't need to change.)

akalpokas:
1. I just tried it again now, and it seems to work for some reason? I'm not sure if the changes I merged from devel fixed it somehow? Regardless, it works now and doesn't trigger circular import.
2. I think it should probably be polished for the next BioSimSpace release, as I basically just copied the code and changed it a bit to work for our inputs. However, for the development version I think it's good to go as it is. I tested it with simple ligand and protein mappings and it all looks good to me.

python/BioSimSpace/Align/_merge.py
```diff
@@ -142,17 +142,9 @@ def merge(
...
-
-        for mol in [molecule0, molecule1]:
-            for idx in roi:
-                if not isinstance(idx, int):
-                    raise TypeError("'roi' must be a list of integers.")
-                if idx < 0 or idx > (mol.nResidues() - 1):
-                    raise ValueError(
-                        f"Residue index {idx} is out of range! The molecule contains {mol.nResidues()} residues."
-                    )
+        from ..Align import _validate_roi
```
lohedges, line 145: I'm actually confused why this works, since `_validate_roi` isn't exposed at the top level in the `Align` submodule, i.e. it's just a private function within the `BioSimSpace.Align._align` module. I would have thought that you want:

```python
from ._align import _validate_roi
```

Indeed, if I do:

```python


In [2]: help(BSS.Align._validate_roi)
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ in <module>:1                                                                                    │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: module 'BioSimSpace.Align' has no attribute '_validate_roi'

In [3]: help(BSS.Align._align._validate_roi)
Help on function _validate_roi in module BioSimSpace.Align._align:

_validate_roi(molecules, roi)
    Internal function to validate that a region of interest (ROI) is a list
    of integers that are within the range of the molecule.

    Parameters
    ----------

    molecules : list
        Consits of a list of :class:`Molecule <BioSimSpace._SireWrappers.Molecule>`.

    roi : list
        The region of interest.
        Consists of a list of ROI residue indices.
```

which is what I would expect.

lohedges: If I run the test after adding the following patch:

```
diff --git a/python/BioSimSpace/Align/_merge.py b/python/BioSimSpace/Align/_merge.py
index 936dc808..c8a9bfa1 100644
--- a/python/BioSimSpace/Align/_merge.py
+++ b/python/BioSimSpace/Align/_merge.py
@@ -141,6 +141,7 @@ def merge(
                 )

     # Validate the region of interest.
+    print("merge", roi)
     if roi is not None:
         from ..Align import _validate_roi

diff --git a/tests/Align/test_align.py b/tests/Align/test_align.py
index a9a9d55c..dfcc6bde 100644
--- a/tests/Align/test_align.py
+++ b/tests/Align/test_align.py
@@ -669,6 +669,9 @@ def test_roi_merge(protein_inputs):
     p0 = BSS.Parameters.ff14SB(p0).getMolecule()
     p1 = BSS.Parameters.ff14SB(p1).getMolecule()

+    print()
+    print("test_roi", roi)
+
     aligned_p0 = BSS.Align.rmsdAlign(p0, p1, roi=roi)
     merged = BSS.Align.merge(aligned_p0, p1, protein_mapping)
     merged_system = merged.toSystem()
```

then I see the following in the output:

```
test_roi [2]
merge None
PASSED
tests/Align/test_align.py::test_roi_merge[double_neighbour_mutant] Downloading from 'https://biosimspace.openbiosim.org/m/double_neighbour_mutant_mut_peptide.pdb'...
Unzipping '/tmp/tmptdso2t4c/double_neighbour_mutant_mut_peptide.pdb.bz2'...
Downloading from 'https://biosimspace.openbiosim.org/m/double_neighbour_mutant_wt_peptide.pdb'...
Unzipping '/tmp/tmpj86a7v_q/double_neighbour_mutant_wt_peptide.pdb.bz2'...

test_roi [2, 3]
merge None
PASSED
```

So the `roi` is correct in the unit test, but is `None` within the merge function itself, which is because the `kwarg` isn't being used in the `merge` call:

```python
    aligned_p0 = BSS.Align.rmsdAlign(p0, p1, roi=roi)
    merged = BSS.Align.merge(aligned_p0, p1, protein_mapping)
    merged_system = merged.toSystem()
```

lohedges: I think the following fixes things:

```
diff --git a/python/BioSimSpace/Align/_merge.py b/python/BioSimSpace/Align/_merge.py
index 936dc808..ba8b6129 100644
--- a/python/BioSimSpace/Align/_merge.py
+++ b/python/BioSimSpace/Align/_merge.py
@@ -142,9 +142,9 @@ def merge(

     # Validate the region of interest.
     if roi is not None:
-        from ..Align import _validate_roi
+        from ._align import _validate_roi

-        _validate_roi(roi, molecule0, molecule1)
+        _validate_roi([molecule0, molecule1], roi)

     # Set 'allow_ring_breaking' and 'allow_ring_size_change' to true if the
     # user has requested to 'force' the merge, i.e. the 'force' argument
diff --git a/tests/Align/test_align.py b/tests/Align/test_align.py
index a9a9d55c..fa6432ed 100644
--- a/tests/Align/test_align.py
+++ b/tests/Align/test_align.py
@@ -670,6 +670,6 @@ def test_roi_merge(protein_inputs):
     p1 = BSS.Parameters.ff14SB(p1).getMolecule()

     aligned_p0 = BSS.Align.rmsdAlign(p0, p1, roi=roi)
-    merged = BSS.Align.merge(aligned_p0, p1, protein_mapping)
+    merged = BSS.Align.merge(aligned_p0, p1, mapping=protein_mapping, roi=roi)
     merged_system = merged.toSystem()
     assert merged_system.nPerturbableMolecules() == 1
```

akalpokas: Ah, that makes sense. It's supposed to have `roi` keyword there, but I forgot to add it. I guess for this small peptide you can use it without it, but if I had a test with a much bigger protein then it would take hours to finish.

akalpokas: If I expose `_validate_roi` in `Align.__init__.py` then I can properly import it in the merge function:
```bash
diff --git a/python/BioSimSpace/Align/__init__.py b/python/BioSimSpace/Align/__init__.py
index 47f97664..c28ed22e 100644
--- a/python/BioSimSpace/Align/__init__.py
+++ b/python/BioSimSpace/Align/__init__.py
@@ -41,3 +41,4 @@ Functions
 """

 from ._align import *
+from ._align import _validate_roi
diff --git a/python/BioSimSpace/Align/_merge.py b/python/BioSimSpace/Align/_merge.py
index 936dc808..8850352b 100644
--- a/python/BioSimSpace/Align/_merge.py
+++ b/python/BioSimSpace/Align/_merge.py
@@ -143,7 +143,6 @@ def merge(
     # Validate the region of interest.
     if roi is not None:
         from ..Align import _validate_roi
-
         _validate_roi(roi, molecule0, molecule1)

     # Set 'allow_ring_breaking' and 'allow_ring_size_change' to true if the
```

akalpokas: Okay, nevermind. It's because I was using `from ..Align import _validate_roi` rather than `from ._align import _validate_roi` like how you're using.

akalpokas: Hopefully that should be to go now!

lohedges: Yes, don't expose it at the `__init__.py` level since this is internal functionality only, so the user shouldn't see it.

lohedges: Looks like it just needs to be blackened, otherwise all should be okay.

akalpokas: Done!

lohedges:
Great. I'll just re-run the CI as a sanity check, then merge tomorrow. Many thanks for this, it's a really nice piece of work. For the next release it would be nice if there was a blog post or tutorial section to highlight this new feature. I'll try to remember closer to the time.

akalpokas:
No worries, thanks for the review! I'm very happy to work on a tutorial for this as there are some peculiarities with the implementation (the way atom ordering needs to be exact between two proteins) for the code to work properly, and I feel like these need to be clearly highlighted with examples in order not to confuse people.

lohedges (approved):

PR #295, opened 2024-05-23
Don't perturb LJ sigma for ghost atoms

lohedges:
This PR switches to not perturbing the LJ sigma parameter for ghost atoms so that merged molecules created by BioSimSpace are compatible with  SOMD2 by default. We still preserve the existing behaviour (zero sigma and epsilon) for SOMD1. The PR also includes a few modifications to harden the checking of dummy atoms during the pert file writing stage. (This isn't needed during merging, since we are explicitly setting the appropriate molecule and atom properties.)

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Align/_merge.py
+++ b/python/BioSimSpace/Align/_merge.py
@@ -693,8 +693,8 @@ def merge(
                     .molecule()
                 )
 
-    # Create a null LJParameter.
-    null_lj = _SireMM.LJParameter()
+    # Tolerance for zero sigma values.
+    null_lj_sigma = 1e-9
 
     # Atoms with zero LJ sigma values need to have their sigma values set to the
     # value from the other end state.
@@ -704,7 +704,7 @@ def merge(
         lj1 = atom.property("LJ1")
 
         # Lambda = 0 state has a zero sigma value.
-        if lj0.sigma() == null_lj.sigma():
+        if abs(lj0.sigma().value()) <= null_lj_sigma:
             # Use the sigma value from the lambda = 1 state.
             edit_mol = (
                 edit_mol.atom(atom.index())
@@ -713,7 +713,7 @@ def merge(
             )
 
         # Lambda = 1 state has a zero sigma value.
-        if lj1.sigma() == null_lj.sigma():
+        if abs(lj1.sigma().value()) <= null_lj_sigma:
             # Use the sigma value from the lambda = 0 state.
             edit_mol = (
                 edit_mol.atom(atom.index())
--- a/python/BioSimSpace/Align/_merge.py
+++ b/python/BioSimSpace/Align/_merge.py
@@ -693,32 +693,31 @@ def merge(
                     .molecule()
                 )
 
-    # Store a dummy element.
-    dummy = _SireMol.Element(0)
+    # Create a null LJParameter.
+    null_lj = _SireMM.LJParameter()
 
-    # Set the LJ sigma paramater to the value of the opposite end state for any
-    # dummy atoms.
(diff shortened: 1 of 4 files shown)


lohedges:
Looks like `ambertype` isn't guaranteed to be present in all molecules, e.g. the alchemical ions from the Exs sandpit. I'll add a check for the property before accessing it tomorrow.

lohedges:
I've fixed the issue with the missing `ambertype` property. There still appears to be an issue using non-zero LJ sigma values for dummy atoms with GROMACS that I am debugging. With the merge code updated to fix the sigma to the opposite end state value, `gmx grompp` now fails when trying to create a binary run file, e.g.:

```
E               RuntimeError: Unable to generate GROMACS binary run input file.
E
E               'gmx grompp' reported the following errors:
E               ERROR 1 [file gromacs.top, line 69]:
E                 Atomtype hcx_du not found
```

I'll check the topology file generated during the test to see what's gone wrong.

chryswoods:
Will do - I will sort that this weekend. I'll change `isDummy` over to testing only the epsilon value (as described in the other thread)

lohedges:
Everything is now working following the latest Sire update.

lohedges:
Just to confirm that I have now checked that the three different approaches for fixing zero LJ sigma values are giving the same result. We now have:

1) The `fix_perturbable_zero_sigmas` map option in `sire` that applies the fix when constructing the perturbable OpenMMMolecule object.
2) A fix in the `somd1` compatibility layer within `somd2`.
3) The updated code in this PR, which ensures that no zero LJ sigma values are present following a merge in BioSimSpace.

The following plot shows a minimised lambda potential energy scan for the ethane --> methanol perturbation. The `sire` and `somd2` results were generated using an _old_ BioSimSpace merged molecule, where zero LJ sigma values were present. The BioSimSpace result just uses the new merging code here, with the fixes from 1) and 2) not applied. I have also generated XML files from OpenMM systems for each data point and have confirmed that (after sorting) they are identical.

zero_sigmas

lohedges:
I'm just wondering about how to preserve the existing `somd1` behaviour with the new approach. If we only fix LJ sigmas for dummy atoms, then everything is easy since it's trivial to detect dummy atoms when the pert file is written. When fixing _all_ zero LJ sigma values, as is done here, then things are more difficult. As an example, for ethane --> methanol I find the following atom record is different when I diff the old and new pert files:

Old:

```
    atom
        name           H4
        initial_type   hc
        final_type     ho
        initial_LJ     2.64953 0.01570
        final_LJ       0.00000 0.00000
        initial_charge 0.03145
        final_charge   0.39600
    endatom
```

New:

```
    atom
        name           H4
        initial_type   hc
        final_type     ho
        initial_LJ     2.64953 0.01570
        final_LJ       2.64953 0.00000
        initial_charge 0.03145
        final_charge   0.39600
    endatom
```

@jmichel80: Are there any issues with this approach? If so, then I could always add another atom property, e.g. `LJ0_somd1` and `LJ1_somd1` that stores specific LJ parameters to use with `somd1` only.

jmichel80:
This would change the behaviour of the softcore I think and could cause changes in the efficiency of transformations with somd1, it would be best to preserve the somd1 protocol.

lohedges:
Here's a comparison of the ethane <--> methanol PMFs for `somd1` with and without the LJ sigma modification, i.e. where the _only_ difference is the change to the non-dummy zero LJ sigma term (for `hc` to `ho`) shown above. These are all short runs and each have unique starting inputs. The overal dG is comparable (within error) and both are reversible,

old_vs_new

chryswoods (approved):
Looks good :-)

PR #302, opened 2024-06-18
Expose torsion modification kwargs

lohedges:
This PR exposes the dummy torsion force constant modification kwargs used by the pert file writer in the `Process.Somd` constructor. These aren't really used, but I want to enable them for debugging purposes. Currently they are set to `False` by default and can only be used if calling the pert file writer function directly.

(This isn't really a bug, since it's not used, but I did forget to expose these when adding kwargs to the process classes to capture engine specific options.)


mb2055 (approved):
All good :+1:

PR #307, opened 2024-07-03

lohedges:
This synchronises with the Exscientia remote ahead of the 2024.2.0 release. Annoyingly it appears that many platform specific skipif directives had been removed from the tests, so I've needed to manually re-add them. Hopefully the CI passes, otherwise I'll need to check what else might have been removed. I'll make a note to mention this since there's no reason to do this. Specific dependencies need to be present for the tests to run. Clearly they are only running their own CI on a specific platform, so thought they could remove them as some minor optimisation.

### DIFF
--- a/.github/workflows/Sandpit_exs.yml
+++ b/.github/workflows/Sandpit_exs.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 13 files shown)


lohedges:
Yes, it looks like I need to add one more `skipif` for GROMACS. How annoying.

PR #313, opened 2024-07-11
Fix issue #312

### LINKED ISSUE #312: [BUG] Unable to merge ions
Related to the discussion [here](https://github.com/OpenBioSim/sire/pull/210), it is currently not possible to merge a water molecule with an ion since the RMSD alignment fails due to there not being enough degrees of freedom. In this case, we should skip the alignment and return an RMSD score of zero for the mapping. This will be useful for creating alchemical waters for use with SOMD2.

lohedges:
This PR fixes #312 by skipping RMSD alignment when merging monatomic ions with other molecules, e.g. water when creating an alchemical ion. When a single item mapping is detected and one of the molecules has a single atom, then we skip the alignment and simply replace the coordinates of the mapped atom with those of the reference. A similar approach could be applied in the mapping/merging code with Sire.

(Note that there is some code duplication below since we still have two scoring functions, one for RDKit mappings, and one for Sire.)

## Suggested reviewers:
@chryswoods

### DIFF
--- a/python/BioSimSpace/Align/_align.py
+++ b/python/BioSimSpace/Align/_align.py
@@ -1547,22 +1547,47 @@ def _rmsdAlign(molecule0, molecule1, mapping=None, property_map0={}, property_ma
     mol0 = molecule0._getSireObject()
     mol1 = molecule1._getSireObject()
 
+    # Do we have a monatomic ion?
+    if (molecule0.nAtoms() == 1) or (molecule1.nAtoms() == 1):
+        is_ion = True
+    else:
+        is_ion = False
+
     # Convert the mapping to AtomIdx key:value pairs.
     sire_mapping = _to_sire_mapping(mapping)
 
     # Perform the alignment, mol0 to mol1.
-    try:
+    if len(mapping) == 1 and is_ion:
+        idx0 = list(mapping.keys())[0]
+        idx1 = list(mapping.values())[0]
+        # Replace the coordinates of the mapped atom with those of the reference.
         mol0 = (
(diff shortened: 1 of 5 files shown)


lohedges:
Bah, it looks like all of the AMBER related tests on Linux have inexplicably started failing. It appears that the `sander.LES` executable is being used, so perhaps this isn't working as expected, or behaves differently for the same input. Will try to debug locally. Amazing that this happens two days after release.

```
FAILED tests/Process/test_amber.py::test_minimise[backbone] - AssertionError: assert not True
 +  where True = <bound method Process.isError of BioSimSpace.Process.Amber(<BioSimSpace.System: nMolecules=631>, BioSimSpace.Protocol.Minimisation(steps=100, restraint='backbone', force_constant=10.0 kcal_per_mol/angstrom**2), exe='$PREFIX/bin/sander.LES', name='test', work_dir='/tmp/tmp5cpac605', seed=None)>()
```

lohedges:
Nope, it's the exact same version of `ambertools`, which matches the version that I have locally:

```
    ambertools:                       23.6-cuda_None_nompi_py310ha112c55_105 conda-forge
```

Everything is fine locally, and for the builds the other day. If a single Python variant failed then I'd think that it was a freak occurrence, but all of them did :man_shrugging:

lohedges:
Looks like the issue is that `sander.LES` is the first executable with `sander` that it's detecting during the CI. I've excluded `LES` from that match. Fingers crossed it works.

lohedges:
Yes, that did the trick. Really confused why this just started happening. Would have been nice if it happened _before_ the release, since it's possible that people creating a fresh environment will experience the problem.

PR #327, opened 2024-08-07
Alchemical Transfer Method

mb2055:
This PR adds functionality for creating, running and analysing RBFE calculations using the Alchemical Transfer Method (AToM).

Functionality is contained in a pair of key locations and is as follows:
* `BioSimSpace.FreeEnergy.AToM` is a class that facilitates the creation of AToM-compatible systems, the use of the BioSimSpace runner to run full sets of production simulations, and analysis using UWHAM.
  * Users can pass a set of three BioSimSpace molecules, and use `...AToM.prepare()` to create an AToM-compatible system. Users can pass a vector to the prepare function that defines the displacement of the free ligand relative to the bound ligand. If this vector is not provided a best-fit vector will be automatically calculated.  
  * The choice of rigid core atoms, which define the alignment of the two ligands and are vital to the stability of simulations, can be tricky. To aid in the choice of these atoms the `AToM` class includes a helper function called `viewRigidCores` which allows the user to view their choice of rigid core atoms within a notebook.
 * `BioSimSpace.Protocol.AToM{}`  contains a series of protocols for minimisation, equilibration and annealing of AToM systems. Equilibration and annealing protocols allow users to manually set the large number of constants and lambda-dependent values used in AToM simulations, any that are not set have sane defaults chosen for them. 
   * `BioSimSpace.Protocol.AToMEquilibration` allows the user to run with or without the ATMForce present, simply by setting `use_atm_force`.
   * `BioSimSpace.Protocol.AToMAnnealing` allows the user to anneal any number of the lambda-dependent values, these are set through a dictionary. If not set the system is annealed from lambda1,lambda2 = 0 to lambda1,lambda2 = 0.5.
   * AToM production protocols are highly sensitive to starting structure, as such its vital that care it taken in the minimisation, equilibration and annealing of  systems. As a rule of thumb the protocol used should be Minimisation->Equilibration without ATMForce -> annealing to lambda1,lambda2=0.5 -> equilibration with ATMForce with lamdba1,lambda2 = 0.5.  Any systems in which instability is experienced in systems with direction=-1 should be addressed by increasing the runtime of post-annealing equilibration.
* Once production simulations have been run (either externally or using `BioSimSpace.FreeEnergy.AToM.run()`) analysis can be performed using `BioSimSpace.FreeEnergy.AToM.analyse()` - this uses an internal UWHAM implementation to find the overall ddg. 

Very roughly, a full start to finish AToM run would look like this:
```python
import BioSimSpace as BSS

lig1 = BSS.IO.readMolecules({ligand1 files})[0]
lig2 = BSS.IO.readMolecules({ligand2 files})[0]
prot = BSS.IO.readMolecules({protein files})[0]

AToM = BSS.FreeEnergy.AToM(protein = prot, ligand1 = lig1, ligand2 = lig2)
AToM_system, data = AToM.prepare(
    ligand1_rigid_core=[a, b, c],
    ligand2_rigid_core=[d, e, f],
    displacement="20A",
)
...
#### solvate AToM_system ####
...
min_protocol = BSS.Protocol.AToMMinimisation(data, {options})
minimise = BSS.Process.OpenMM(AToM_system, min_protocol)
minimise.start()
minimise.wait()
minimised = minimise.getSystem()
...
#### now do equilibration, annealing and ATM equilibration, API same as minimisation, 
#### setting the required options) ####
...
production_atm = BSS.Protocol.AToMProduction(
    data=data,
    {options}
)
production = AToM.run(
    {minimised and equilibrated AToM_system},
    production_atm,
    work_dir = {output directory},
)
production.wait()

BSS.FreeEnergy.Analyse({output directory})
```

This PR also includes (hopefully) full API docs, as well as a brief summary of AToM within the docs for `BioSimSpace.FreeEnergy`.

--------------------------------
* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): [y/n]
* I confirm that I have added a test for any new functionality in this pull request: [y/n]
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [y/n]
* I confirm that I have permission to release this code under the GPL3 license: [y/n]

## Suggested reviewers:
@lohedges


python/BioSimSpace/Protocol/_AToM.py
```diff
@@ -0,0 +1,3287 @@
+from BioSimSpace.Units.Area import angstrom2
```
lohedges, line 1: Could this be called `_atm.py` to match the modules in the other sub-modules, e.g. `Process._atm` etc.

python/BioSimSpace/Process/_atom_utils.py
```diff
@@ -0,0 +1,977 @@
+######################################################################
```
lohedges, line 1: Can this file be call `_atm_utils.py` to match the other modules.

python/BioSimSpace/Process/_openmm.py
```diff
@@ -68,6 +67,34 @@ class OpenMM(_process.Process):
...
+    ):
+        from ._atm import OpenMMAToM
+
+        # would like to use issubclass but _Protocol._AToM is not exposed
+        if (
+            isinstance(protocol, _Protocol.AToMMinimisation)
+            or isinstance(protocol, _Protocol.AToMEquilibration)
+            or isinstance(protocol, _Protocol.AToMAnnealing)
+            or isinstance(protocol, _Protocol.AToMProduction)
+        ):
```
lohedges, line 92: While the base `_AToM` protocol isn't exposed via the public API, you should just be able to import and use it directly, e.g.:

```python
from ..Protocol._atm import _AToM

if isinstance(protocol, _AToM):
    ...
```

Note that `isinstance` will work because this is an actual instance, rather than a class.

python/BioSimSpace/Protocol/_AToM.py
```diff
@@ -0,0 +1,3287 @@
...
+
+        Returns
+        -------
+
+        data : dict
+            The AToM data dictionary.
+        """
+        return self._system_data
+
+    def _getCoreAlignment(self):
```
lohedges, line 148: Can the setters and getters be public, as is the case for the other protocols? (That is, don't start with an `_`.) The use case was that sometimes protocols need to get modified further down the chain, or by the user, so having exposed setters (that validate things in a self-consistent way) makes it easier to avoid the user breaking things, e.g. by trying to set the attribute directly. I imagine that the AToM functionality is pretty self contained, so this is unlikely needed,  but it would be good to make things consistent.

python/BioSimSpace/Protocol/_AToM.py
```diff
@@ -0,0 +1,3287 @@
...
+            The uh value for the ATM softcore potential (kcal/mol).
+
+        SC_a : int, float, str, :class:`Energy <BioSimSpace.Types.Energy>`
+            The a value for the ATM softcore potential.
+
+        cm_kf : int, float, str, :class:`GeneralUnit <BioSimSpace.Types._GeneralUnit>`
+            The force constant for the center of mass distance restraint (kcal/mol/A^2).
+
+        cm_tol : int, float, str, :class:`Length <BioSimSpace.Types.Length>`
+            The tolerance for the center of mass distance restraint (A)."""
```
lohedges, line 754: I'm guessing these might map one-to-one with the original Gallicchio implementation, but it might be nice to make the meaning of the kwargs a little clearer and the formatting standardised. For example, I'd prefer if everything was lower snake_case. Things like `core_alignment` seem self explanatory, whereas `CMCM_restraint` might be clearer as `com_distance_restraint` (along with related `com_force_constant` and `com_tolerance`). There are also mixtures of formatting, e.g. `rest`, `restraint`, `restrained`. These could al be `restraint` for consistency.

The latter kwargs are a little confusing if you aren't familiar with the method. (Specifically the `SC` related terms.) I guess these probably will take the default values in most instances, so the user won't really be bothered by them. I'd still prefer if the names and description were a little more informative, though. Perhaps you could add a `Notes` section to the start of the constructor docstring to link to the AToM expressions, or write them if formatting allows? (This might be easier in website API docs.)

There are few inconsistencies in the kwargs, such as both `k` and `kf` being used for force constant, as well as `force_constant` by itself. I'd prefer the same througout, with `force_constant` making things clear. `SC` could be `soft_core` for clarity.

There are a few minor typos too, e.g. `restaint` instead of `restraint`.

python/BioSimSpace/Process/_atom_utils.py
```diff
@@ -0,0 +1,977 @@
...
+        # Check for proper typing
+        if not isinstance(
+            protocol,
+            (
+                _Protocol.AToMMinimisation,
+                _Protocol.AToMEquilibration,
+                _Protocol.AToMAnnealing,
+                _Protocol.AToMProduction,
+            ),
+        ):
```
lohedges, line 40: This could also use:

```python
from ..Protocol._atm import _AToM

if not isinstance(_AToM):
    ...
```

python/BioSimSpace/Process/_atom_utils.py
```diff
@@ -0,0 +1,977 @@
...
+        self.findDisplacement()
+        d = [round(x, 3) for x in self.displacement]
+        output = ""
+        output += "displacement = {}\n".format(d)
+        output += "#BioSimSpace output is in angstrom, divide by 10 to convert to the expected units of nm\n"
+        output += "displacement = [i/10.0 for i in displacement]\n"
+        return output
+
+    def createSoftcorePertE(self):
+        """Create the softcorePertE function for the Gallachio lab analysis"""
```
lohedges, line 153: It's Gallicchio. This will need to be changed throughout.

python/BioSimSpace/FreeEnergy/_ddg.py
```diff
@@ -0,0 +1,495 @@
...
+import pandas as _pd
+
+import numpy
+import scipy.optimize
+import scipy.special
+import pathlib
+import functools
+import pathlib as _pathlib
+import os as _os
+import pymbar
```
lohedges, line 36: Can these all be private.

python/BioSimSpace/FreeEnergy/_ddg.py
```diff
@@ -0,0 +1,495 @@
...
+        The point at which 'direction' changes.
+        Should be (last index of direction 1, first index of direction 2).
+        If not provided not provided, will be implied from files.
+
+    Returns
+    -------
+    ddg_total : float
+        The free energy.
+    ddg_total_error : float
+        The error in the free energy.
```
lohedges, line 221: Can these be `BioSimSpace.Types.Energy` types.

python/BioSimSpace/FreeEnergy/_atm.py
```diff
@@ -0,0 +1,1393 @@
...
+            The indices at which the direction changes. For example, if direction=[1,1,-1,-1],
+            then inflex_indices=[1,2].
+            If None, the inflexion point will be found automatically.
+
+        Returns
+        -------
+        ddg : float
+            The free energy difference between the two ligands.
+        ddg_err : float
+            The error in the free energy difference.
```
lohedges, line 1092: Return `BioSimSpace.Types.Energy` types.  I guess this might also need to return a overlap matrix, conditional on the `method`?

python/BioSimSpace/FreeEnergy/_atm.py
```diff
@@ -0,0 +1,1393 @@
...
+from .._SireWrappers import Molecule as _Molecule
+from .._SireWrappers import System as _System
+from .. import _Utils
+from ..Types import Length as _Length
+from ..Types import Vector as _Vector
+from ..Types import Coordinate as _Coordinate
+from ..Align import matchAtoms as _matchAtoms
+from ..Align import rmsdAlign as _rmsdAlign
+from ..Notebook import View as _View
+from ..Process import OpenMM as OMMprocess
```
lohedges, line 36: Can this be private, e.g. `from ..Process import OpenMM as _OpenMM`.

python/BioSimSpace/FreeEnergy/_atm.py
```diff
@@ -0,0 +1,1393 @@
...
+
+    @staticmethod
+    def run(
+        system,
+        protocol,
+        platform="CPU",
+        work_dir=None,
+        setup_only=False,
+        property_map={},
+    ):
```
lohedges, line 1020: This is behaving like the constructor of `BioSImSpace.FreeEnergy.Relative`. (It basically constructs a similar object, then calls `.run` on it.) I'll have a think about how best to do this. Here `BioSimSpace.FreeEnregy.AToM` basically does all the setup, whereas the additional `BioSImSpace.FreeEnergy._relativeATM` (which is created by this `run` method) runs the actual simulation.

If possible, it might be better to have `BioSimSpace.AToM.prepare` and `BioSImSpace.AToM.viewRigidCores` as standalone functions, then `BoSimSpace.FreeEnergy.AToM` as the main class that can setup, run, and analyse the simulation, as is done for `BioSimSpace.FreeEnergy.Relative. I understand that `BioSimSpace.FreeEnergy.AToM` stores a lot of state, which simplifies the call to `.prepare`, etc. It might be possible to just change the class to `AToMSetup`, or just `Setup`, to make clear what it's being used for. Will have a think.

python/BioSimSpace/Protocol/_AToM.py
mb2055: This is now updated (https://github.com/OpenBioSim/biosimspace/blob/feature_AToM/python/BioSimSpace/Protocol/_atm.py)

python/BioSimSpace/Process/_atom_utils.py
mb2055: Now changed (https://github.com/OpenBioSim/biosimspace/blob/feature_AToM/python/BioSimSpace/Process/_atm_utils.py)

python/BioSimSpace/Process/_openmm.py
mb2055: This has been updated accordingly (https://github.com/OpenBioSim/biosimspace/blob/feature_AToM/python/BioSimSpace/Process/_openmm.py#L88)

python/BioSimSpace/Protocol/_AToM.py
mb2055: Setters and getters are now public. There are two exceptions to this:
- `_set_current_index` and `_get_current_index` in the annealing protocol. This is used internally to set the initial values of the lambda-dependent variables, which should not be touched by the user. 
- `_set_lambda_values` and `_get_lambda_values` within the production protocol. These are used to make the _master_ lambda value, but don't change anything internally (they're just for file naming), allowing the user to set these would just cause confusion. It should also be noted that the production protocol has its own versions of `set_current_index` and `get_current_index`. These are something that could very easily break things should the user not set them properly, and they might need to be set to private.

python/BioSimSpace/Protocol/_AToM.py
mb2055: Kwargs have now been updated to (hopefully) make things clearer. The changes are as follows:
- `CMCM_restraint` -> `com_distance_restraint`
- `cm_kf` -> `com_k`
- `cm_tol` -> `com_restraint_width`
- `align_kf_sep` -> `align_k_distance`
- `pos_rest_width` -> `positional_restraint_width`
- `SC_umax` -> `soft_core_umax`
- `SC_u0` -> `soft_core_u0`
- `SC_a` -> `soft_core_a`
The constants have also been grouped to make it clearer which goes with which (e.g. `com_k` now directly follows `com_distance_restraint` in the arguments)

python/BioSimSpace/Process/_atom_utils.py
mb2055: Now updated (https://github.com/OpenBioSim/biosimspace/blob/feature_AToM/python/BioSimSpace/Process/_atm_utils.py#L33)

python/BioSimSpace/Process/_atom_utils.py
mb2055: Now changed throughout. Hopefully I didn't miss any.

python/BioSimSpace/FreeEnergy/_ddg.py
mb2055: Now private (https://github.com/OpenBioSim/biosimspace/blob/feature_AToM/python/BioSimSpace/FreeEnergy/_ddg.py#L27-L35)

python/BioSimSpace/FreeEnergy/_ddg.py
mb2055: ddg now returns energies

python/BioSimSpace/FreeEnergy/_atm.py
mb2055: As it stands, since `UWHAM` is the only functioning analysis, I've not added any additional returns from `analysis`, if we are able to get `MBAR` working in the future this can be revisited

python/BioSimSpace/FreeEnergy/_atm.py
mb2055: Now private (https://github.com/OpenBioSim/biosimspace/blob/feature_AToM/python/BioSimSpace/FreeEnergy/_atm.py#L36)

lohedges:
Thanks, after refactoring this all looks good. I'll deal with the Sire release this week then give this another once over before doing the BioSimSpace side of things.

lohedges (approved):
Many thanks for this heroic effort, @mb2055. Things are looking really good and the new tutorial is great.

PR #331, opened 2024-08-16
Fix issue #330

### LINKED ISSUE #330: [BUG] BSS.Convert.toRDKit() appears to lose stereochemical information
Hey again,

I think I've encountered a bug with the Convert module (or I might be overlooking a crucial bit of code). Converting from a BioSimSpace / Sire mol to an rdmol via BSS.Convert.toRDKit() appears to lose stereochemical info.

**To Reproduce**

```py
from rdkit import Chem
import BioSimSpace as BSS

# Define enantiomer
a = 'CC[C@@H](O)C'

# Check stereochemistry of the molecules
a

# Parameterise
bs_a = BSS.Parameters.gaff(a).getMolecule()

# Check the stereochemistry is retained in the parameterised molecules 
bs_a._sire_object.view()

# Convert back to rdkit
rd_a = BSS.Convert.toRDKit(bs_a)
rd_a
```

**Expected behavior**
It seems to me like there is no reason for the stereochemistry to be removed.  The reason this has become an issue for me is because I'm using an rdkit-based flexible alignment method rather than rmsd/rmsdAlign, and so working with rdkit molecules that exhibit correct stereochemical centres is important. Perhaps I've missed something in the docs that'll help me achieve this. 

OS: [e.g. Linux (Ubuntu 20.04.6 LTS (Focal Fossa)
Version of Python: Python 3.12.4
Version of BioSimSpace: 2024.2.0
I confirm that I have checked this bug still exists in the latest released version of BioSimSpace: yes

Cheers,
Noah

lohedges:
This PR closes #330 by preserving SMILES based molecule properties when parameterising. Previously, `BioSimSpace.Convert.smiles` was only used to create an intermediate molecule during parameterisation, so any properties that were generated at this point were _not_ part of the returned molecule. This PR fixes this by adding them to the molecule, hence preserving attributes that are required for stereochemistry.

## Suggested reviewers:
@mb2055

### DIFF
--- a/python/BioSimSpace/Parameters/_Protocol/_amber.py
+++ b/python/BioSimSpace/Parameters/_Protocol/_amber.py
@@ -310,15 +310,24 @@ class AmberProtein(_protocol.Protocol):
         if work_dir is None:
             work_dir = _os.getcwd()
 
-        # Flag whether the molecule is a SMILES string.
+        # Try to create a molecule from the SMILES string.
         if isinstance(molecule, str):
-            is_smiles = True
-        else:
-            is_smiles = False
+            try:
+                smiles = molecule
+                molecule = _smiles(molecule)
+                edit_mol = molecule._sire_object.edit()
+                edit_mol = edit_mol.rename(f"smiles:{smiles}").molecule()
+                molecule._sire_object = edit_mol.commit()
+            except Exception as e:
+                msg = "Unable to convert SMILES to Molecule using RDKit."
+                if _isVerbose():
+                    msg += ": " + getattr(e, "message", repr(e))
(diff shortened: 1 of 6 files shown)


lohedges:
Looks like this doesn't completely fix #330, rather the specific case of SMILES input. Note that the test that fails is passing locally. Will investigate further.

lohedges:
Okay, it works locally using the same environment as for the CI. The test involves generating a mapping with RDKit, so I had assumed that this had changed between versions. (It doesn't specify a conformer to use.) If it had failed on a single platform I would have put it down to bad luck, but here it's failed on all of them. (The test is skipped on Windows.)

PR #333, opened 2024-09-06
Fix issue #332

### LINKED ISSUE #332: [BUG] Unable to align based on single atom mappings
Related to #312, it is not possible to align any single atom mapping. Previously we added a fix for ions, but this should be relaxed to allow all single atom mappings, whether the mapped atom contains an ion or not.

lohedges:
This PR closes #332 by relaxing the constraint on the alignment of single atom mappings, i.e. they no longer need to involve an ion. The original bug was caused by the mapping having too few degrees of freedom, not the molecule itself.


mb2055 (approved):
Looks good :+1:

PR #336, opened 2024-09-13
Fix issue #335

### LINKED ISSUE #335: [BUG] system.updateMolecules() doesn't work correctly for multiple molecules
Currently the `system.updateMolecules` method doesn't work correctly when multiple molecules are passed in a list since the underlying Sire object (`system._sire_object`) is only udpated after the final molecule has been updated, not at each iteration. This breaks the logic since a copy of the system is used within the loop to avoid the original system being destroyed if Sire throws an exception on update.

lohedges:
This PR fixes #335 by ensuring that the underlying Sire object of a BioSimSpace system is updated during the inner loop when updating molecules within the system. Otherwise, only the final molecule will be updated.


mb2055 (approved):
Looks good :+1:

PR #338, opened 2024-09-20
Add method to remove the space property from a system

lohedges:
This PR adds a method to `BioSimSpace._SireWrappers.System` to remove the space property from a system. At present this doesn't warn the user if there is more than one molecule, e.g. it's not a vacuum system. I'm happy to add this if useful. (I imagine the use case here is removing water from a box, then removing the space.)


mb2055 (approved):
All good :+1:

PR #345, opened 2024-09-27
Fix issue #344

### LINKED ISSUE #344: [BUG] Merge code no longer setting a unique connectivty property when connectivity changes
It looks like this happened during the merging of the ROI code, where presumably the sandpit code was taken verbatim rather than being adapted to add the new functionality. We require a _unique_ connectivity object in the end states when the connectivity changes, as can be done when forcing a merge.

lohedges:
This PR fixes #344 by reverting the way in which the connectivity property is set. While we generally handle perturbations that _don't_ change connectivity, we do require that the end state properties differ, These properties will be used by Sire to infer situations where the perturbation involves a connectivity change.

### DIFF
--- a/python/BioSimSpace/Align/_merge.py
+++ b/python/BioSimSpace/Align/_merge.py
@@ -1162,8 +1162,13 @@ def merge(
                             "perturbation will likely be unstable."
                         )
 
-    # Set the "connectivity" property.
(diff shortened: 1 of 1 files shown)


mb2055 (approved):
All good :+1:

PR #349, opened 2024-10-01
Fix issue #348

### LINKED ISSUE #348: [BUG] Use direct to/from RDKit conversion when parameterising with OpenFF
Going via and RDKit format intermediate should avoid issues with using file formats, e.g. PDB/SDF.

lohedges:
This PR closes #348 by using direct conversion to/from and RDKit format intermediate for parameterisation with OpenFF. This avoids issues using intermediate file formats, e.g. PDB or SDF. We first convert from SIre/BioSimSpace to RDKit, then create an OpenFF molecule directly from this using the `Molecule.from_rdkit()` functionality.


mb2055 (approved):
All good

Issue #359, opened 2024-10-28
ABFE restraints torsional force constants are too high


fjclark:
**Describe the bug**

This is a fairly minor, but I've found that my restraints fitting code is selecting torsional force constants ~ 3 times stronger than is consistent with normal fluctuations in the unrestrained state. This does not affect the correctness of any ABFE calculations (e.g. the restraint corrections are still correct and lower torsional force constants will not affect stability).

I came across this when I realised that I could test the restraint search code by feeding it a trajectory of a completely decoupled state (with some Boresch restraints applied), and the restraint code should recover exactly the restraints used to simulate the decoupled state. I found that this worked, other than the force constants being ~ 3 times too large. The issue is [here](https://github.com/OpenBioSim/biosimspace/blob/eb44a6fde5c2261afa6758b10f5afeacf45ad33a/python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py#L1491):

```
                        for val in boresch_dof_data[pair][dof]["values"]:
                            dtheta = abs(val - circmean)
                            corrected_values.append(min(dtheta, 2 * _np.pi - dtheta))
                        corrected_values = _np.array(corrected_values)
                        boresch_dof_data[pair][dof]["var"] = corrected_values.var()
```

Here, I'm shifting the deviations between the measured torsions and the mean value so that they are in the range [-pi, pi], consistent with their periodicity. However, I've forgotten that because I've taken the absolute value, the deviations are all on the same side of the mean, and so `corrected_values.var()` on the last line is incorrect. Instead, it should be `np.mean(corrected_values**2)`, which avoids erroneously re-centring through subtracting the mean of the deviations.

Assuming that the deviations in the unrestrained state used to fit the restraints are normal (which has been a pretty good approximation in my experience, at least for well-behaved systems), then this bug produces an overestimate of the variance by a factor of ~3:
```
>>> data = np.random.normal(size=1_000_000, scale=1)
>>> 1/np.var(abs(data))
2.75
```
(whereas `1/np.mean(abs(data)**2) = 1`)

Boresch restraints are subject to instabilities if the angular force constants are too low, but lowering the torsional force constants should not affect stability. 

I'll fix this along with the other discussed tweaks to the restraint selection code.

PR #360, opened 2024-10-28
Improve robustness of restraint search and fix torsional force constant bug

fjclark:

This PR does two things:

1. Fixes https://github.com/OpenBioSim/biosimspace/issues/359
2. Improves the robustness of the ABFE restraint search to strange user input. I've implemented the improvements discussed [here](https://github.com/OpenBioSim/biosimspace_tutorials/issues/28#issuecomment-2434686918). 

I've included the two together to reduce CI runs/ faff, but please let me know if I should create separate PRs in future.

## If this is to fix a bug...

This pull request fixes issue #359?

Note that the ordering of candidate anchor points in the restraint search will be slightly different now, as the torsions have less weight when ordering the restraints for selection.

I now:

 - Ensure that anchors can never be in the same molecule
 - Warn the user if they select the protein/water as the "ligand"
 - Warn the user if they supply very few frames for fitting

* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [n]

The tests could be more comprehensive - at the moment I've only added a test to check that an error is raised if the receptor and ligand candidate anchor point lists share atoms. I haven't included tests to check that the correct warnings are being raised. The reason for this was that the only trajectory I can find on the website (https://github.com/OpenBioSim/biosimspace_website/tree/main/structures) only includes the ligand and the protein. This means that I can't select weird "ligands" which aren't the ligand or protein (e.g. water) which don't raise errors for other reasons. If we'd like more complete tests, we could always add a trajectory with e.g. 1 water molecule as well as the complex. However, I have verified that these warnings are raised for one of my test systems, and the code for these checks is very simple.

## Suggested reviewers:
@lohedges 

Thanks.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -63,7 +63,7 @@ Tutorials
 =========
 
 .. toctree::
-   :maxdepth: 1
+   :maxdepth: 2
 
    tutorials/index
 
@@ -71,7 +71,7 @@ Detailed Guides
 ===============
 
 .. toctree::
-   :maxdepth: 1
+   :maxdepth: 2
 
    guides/index
 
@@ -93,7 +93,7 @@ Support
 Contributing
 ============
 .. toctree::
-   :maxdepth: 1
+   :maxdepth: 2
 
    contributing/index
    contributors
--- a/python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
@@ -590,6 +590,14 @@ class RestraintSearch:
                 f"traj {type(traj)} must be of type 'BioSimSpace.Trajectory._trajectory.Trajectory'"
             )
 
+        n_frames = traj.nFrames()
+        if not n_frames >= 50:
+            _warnings.warn(
+                f"The trajectory for restraint selection has less than 50 frames ({n_frames} frames). "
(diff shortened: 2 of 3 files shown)


lohedges:
Thanks for this. The CI is failing because there isn't yet a 2024.4.0.dev Sire package. I'll trigger the CI for that now.

lohedges:
Not sure what's going on, but it appears that you might have merged in `devel` without actually taking the changes, i.e. using `ours`. This is why the CI is failing, since you are still building against the old version of Sire. (Although the packages should still be available.) For example, [this](https://github.com/fjclark/biosimspace-openbiosim/blob/feature-robust-restraint-search/requirements.txt) requirements files appears to be the one from `main`, not `devel`. I think this might also explain the weird diff with the website index file, which is proposing changes which are _already_ in `devel`.

fjclark:
Sorry, not sure what I've done to cause this. Double-checking that my `devel` was in sync with OBS, pulling again, and merging again changed only `requirements.txt`:

image

lohedges:
No problem. Things look correct now. Ignore the Windows failure, although the CI passed, it crashed when uploading the package to the Anaconda cloud. The Linux 3.12 failure is due to black formatting. (I need to switch to pre-commit at some point so the linting/formatting is automated.)

lohedges (approved):
Thanks, @fjclark. This looks great and will make the restraint selection a lot more reliable.

PR #361, opened 2024-10-28
Fix issue #347

### LINKED ISSUE #347: Negatively charged nitrogen parameterisation
Hi,

I've been encountering an issue with antechamber recently - it seems like it really doesn't like negatively charged nitrogens. [Here is an example molecule in SDF format.](https://drive.google.com/file/d/13zFTDpAJp0rxIQuSG3YQIfB3Zt0tITi_/view?usp=sharing) This is a simplified version since I can't share the structures I'm working with.

Antechamber error:
> antechamber: Fatal Error!
> Weird atomic valence (2) for atom (ID: 40, Name: N8).
>        Possible open valence.

Is there anyway for me to continue using the interface BSS provides to antechamber while ignoring this valence error? I see it's possible to use the -dr no flag, which I assume avoids running the valence checks (although i'm struggling to find documentation for this) but I'm not sure if there's a way to add this to the command that BSS generates. I feel like if the molecule is 'sanitised' by BSS/rdkit then we don't necessarily need to worry about antechamber running its own valence checks, perhaps.

I've also been looking into using OpenFF instead of gaff2 but their documentation seems to suggest you should use the 'constrained' forcefield for any MD over 1fs timesteps. Since I'm using a timestep of 3fs I suppose I should avoid using the openff forcefields implemented via BSS? Are there any plans to include the constrained forcefields in the future?

Thanks,
Noah

lohedges:
This PR closes #347 by allowing the Antechamber `acdoctor` mode to be disabled via a kwarg when using `BioSimSpace.Parameters.gaff/gaff2`, or via `BioSimSpace.Parameters.parameterise`.

I'll re-trigger the CI once the Sire development packages have finished building.

@noahharrison64: Tagging you in so that you can test.

### DIFF
--- a/python/BioSimSpace/Parameters/_Protocol/_amber.py
+++ b/python/BioSimSpace/Parameters/_Protocol/_amber.py
@@ -841,6 +841,7 @@ class GAFF(_protocol.Protocol):
         net_charge=None,
         ensure_compatible=True,
         property_map={},
+        **kwargs,
     ):
         """
         Constructor.
@@ -873,6 +874,10 @@ class GAFF(_protocol.Protocol):
             A dictionary that maps system "properties" to their user defined
             values. This allows the user to refer to properties with their
             own naming scheme, e.g. { "charge" : "my-charge" }
+
+        **kwargs: dict
+            Additional keyword arguments. These can be used to pass custom
+            parameters to the Antechamber program.
         """
 
         if type(version) is not int:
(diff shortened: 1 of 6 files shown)


noahharrison64:
Can confirm this branch allows the successful parsing of acdoctor = False as a keyword arg to the gaff2 parameterisation function. Doing so sets -dr no in the antechamber command and allows for parameterisation of molecules with unusual valences that antechamber may not be set up to handle (negatively charged nitrogens in my use case)

Thanks @lohedges

PR #366, opened 2024-11-13
Fix AMBER FEP mappings

lohedges:
This PR adds the missing mapping pruning functions from the Exscientia, which are required for stable mappings when using AMBER as the FEP engine. This functionality is tested within the sandpit. I've also tidied a few bits of the mapping code and made sure that sub-functions take kwargs so that we don't get bugs if the order of inputs changes in future.


lohedges:
The PR also includes edits that are required for @annamherz's RBFE pipeline code.

mb2055 (approved):
Looks good :+1:

PR #367, opened 2024-11-13
Fix amber fep mappings - protocol updates

annamherz:
This PR is for further updates relating to running free energy simulations, copying settings from [this branch](https://github.com/michellab/BioSimSpace/releases/tag/RBFE-benchmark).

Changes:
-  Updates to the FreeEnergy for data processing. This includes [concat outside of the function](https://github.com/annamherz/biosimspace/blob/5dd36b79ff3c795b43171b5896f3c408fefb7d6c/python/BioSimSpace/FreeEnergy/_relative.py#L1311), so that the preprocessing function can be used independently for other analysis outside of BSS. I also updated the [truncate arguments](https://github.com/annamherz/biosimspace/blob/5dd36b79ff3c795b43171b5896f3c408fefb7d6c/python/BioSimSpace/FreeEnergy/_relative.py#L1161-L1162) for more clarity.
- All steps for reporting to 200 (from 250). This matches what is in the default protocol settings.
- 10 A cutoff for PME and 12 A for reaction field.
- Gromacs protocol softcore settings update for better default values, as well as some additional detail for the integrators/barostats.

Questions:
- In the process for Gromacs, only full perturbations [are allowed](https://github.com/annamherz/biosimspace/blob/5dd36b79ff3c795b43171b5896f3c408fefb7d6c/python/BioSimSpace/Process/_gromacs.py#L265-L274). This is also the case [previously for Amber in the free energy function](https://github.com/michellab/BioSimSpace/blob/0c14a0c4e470f5226c44875af2eb78e895b99969/python/BioSimSpace/FreeEnergy/_relative.py#L296C1-L300C1). Does this need to be added to the Amber process?

## Suggested reviewers:
@lohedges

### DIFF
--- a/python/BioSimSpace/Protocol/_equilibration.py
+++ b/python/BioSimSpace/Protocol/_equilibration.py
@@ -50,6 +50,7 @@ class Equilibration(_Protocol, _PositionRestraintMixin):
         thermostat_time_constant=_Types.Time(1, "picosecond"),
         report_interval=200,
         restart_interval=1000,
+        restart=False,
         restraint=None,
         force_constant=10 * _Units.Energy.kcal_per_mol / _Units.Area.angstrom2,
     ):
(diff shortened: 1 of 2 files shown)


lohedges:
Thanks for this. The tweaks are now incorporated in to #366, so we can close this when you're happy.

PR #368, opened 2024-11-15
Improve error messages if no lambda windows found during analysis

fjclark:

If users have generated output without using BSS, the directory structure may not be as BSS expects when parsing lambda windows from directory names. Previously, this lead to a an empty `lambdas` list, and `BSS.FreeEnergy.Analyse` returned `([], correct_overlap_matrix)` without raising an error. The code now explicitly raises an error.

* I confirm that I have added a test for any new functionality in this pull request: n

## Suggested reviewers:
@lohedges

### DIFF
--- a/python/BioSimSpace/FreeEnergy/_relative.py
+++ b/python/BioSimSpace/FreeEnergy/_relative.py
@@ -1435,6 +1435,12 @@ class Relative:
             for part in file.parts:
                 if "lambda" in part:
                     lambdas.append(float(part.split("_")[-1]))
+        if len(lambdas) == 0:
+            raise ValueError(
+                "No lambda windows were detected from the output directory names! "
+                "Ensure that the directory being analysed contains sub-directories"
+                "name e.g `lambda_0`, `lambda_1` containing the output files."
+            )
 
         # Find the temperature for each lambda window.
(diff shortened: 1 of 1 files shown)


lohedges (approved):
Thanks, @fjclark. I think the directory naming convention here refers to what is used in the sandpit rather than the core code, but is clear enough. (I plan to align this anyway.)

lohedges:
Hmm, this somehow broke the code. (CI failed on all platforms.) Just debugging now.

lohedges:
Ah yes, indentation error. Will fix now.

lohedges:
I've now reworked this. There isn't the need to have the error in multiple places. I've just put it in the internal anaysis method so it works for all engines, e.g. `somd2` was missing from the previous edits.

fjclark:
Apologies for the rushed and sloppy PR. I'll be more careful about testing in the future. I didn't add this check for SOMD2 because it doesn't expect a directory for each lambda as with the other engines, and it seemed much less likely that the output could be formatted in a way BSS doesn't expect.

lohedges:
No problem. It was obvious what was wrong when I looked at the whole file rather than the diff :-)

PR #379, opened 2025-01-07
Fix issue #376

### LINKED ISSUE #376: RMSD reference containing protein and ligand
**Is your feature request related to a problem? Please describe.**
I want to use BioSimSpace.Metadynamics.CollectiveVariable.RMSD on a ligand pose

**Describe the solution you'd like**
I would like to be able to use BioSimSpace.Metadynamics.CollectiveVariable.RMSD with a ligand pose in its pocket which means a reference containing more than one molecule

**Describe alternatives you've considered**
I don't see an alternative to using a multi-molecule reference when using RMSD CVs with ligand poses

**Additional context**

lohedges:
This PR closes #376 by allowing the user to specify unique atom selections to be used for the RMSD and alignment when using an RMSD restraint. This means that you can use one selection to align to, and another for the calculation of the RMSD itself. The PR also aligns some small differences between the core and Sandpit code bases. (Some minor updates must not have been duplicated.)

Note that this PR creates a necessary API change, so we will need to update the steered MD [tutorial](https://github.com/OpenBioSim/biosimspace_tutorials/tree/main/03_steered_md) accordingly when we do the next release.

Tagging in @AdeleHardie so that you are aware of this change. The updated [unit test](https://github.com/OpenBioSim/biosimspace/blob/95da10af2ae66da23e33a16d1ce82233d64285a9/tests/Metadynamics/test_metadynamics.py#L47) shows how the new API used.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -43,7 +43,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "biosimspace"
-copyright = "2017-2024"
+copyright = "2017-2025"
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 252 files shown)


mb2055 (approved):
All good :+1: , sorry for the delay

PR #385, opened 2025-01-27
Fix issues 381, 382, and add a Mass type

### LINKED ISSUE #381: [BUG] GeneralUnit has no _to_default_unit() method
Since, by definitiion, `GeneralUnit` is already in default internal units, there is no `_to_default_unit()` method. This causes issues when a `GeneralUnit` instance is treated like the other unit-based types. An easy fix is to just add a `_to_default_unit()` method that returns `self`, so it can be treated the same way.

### LINKED ISSUE #382: Refined mapping of perturbable regions for AMBER input files
Hello! When mapping molecules to each other for AFE simulations, AMBER experiences issues when running simulations where a perturbable part of the molecule separates two shared regions of the molecule. This causes the following error message:
image
All of these simulations were tested at 4 fs with HMR applied, but the error also persisted at 2 fs.

An example of such a mapping is (lig_2n~lig_2o):
image

But also in another case, where ring breaking was permitted so that the mapping was able to proceed:(lig_CHEMBL3265002~lig_CHEMBL3265004):
image

In all of these cases, there is a shared atom separated from the rest of the molecule. In some cases, additionally pruning the atoms using the `prune_crossing_constraints` is able to overcome this (lig_2n~lig_2o):
image

but there are other cases where the atoms remain in the common atoms and the crash then still occurs (lig_CHEMBL3265002~lig_CHEMBL3265004):
image

Here are sample files for both: [example_ligands.zip](https://github.com/user-attachments/files/18414108/example_ligands.zip)

I think a potential solution could be to check that all atoms that are in the common core are connected for the AMBER files, and if they are not connected to the largest common core, to remove them from the shared mapping (in the cfg file) and designate them as perturbable?

lohedges:
This PR closes #381 by adding a `._to_default_unit()` method to the internal `GeneralUnit` class. This simply returns `self`, since they are by definition in internal default units.

The PR also closes #382 by clarifying the use of the `explicit_dummies` kwarg in the `Process.Amber` constructor docstring.

The PR also adds support for a new `Mass` type, a `Mass` requirement for nodes, and `Mass` unit convenience classes.

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/Types/_mass.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/Types/_mass.py
@@ -1,0 +1,402 @@
+######################################################################
+# BioSimSpace: Making biomolecular simulation a breeze!
+#
+# Copyright: 2017-2025
+#
+# Authors: Lester Hedges <lester.hedges@gmail.com>
+#
+# BioSimSpace is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# BioSimSpace is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with BioSimSpace. If not, see <http://www.gnu.org/licenses/>.
+#####################################################################
+
+"""A mass type."""
+
+__author__ = "Lester Hedges"
+__email__ = "lester.hedges@gmail.com"
+
+__all__ = ["Mass"]
+
+from sire.legacy import Units as _SireUnits
+
+from ._type import Type as _Type
+
+
+class Mass(_Type):
+    """A mass type."""
+
+    # A list of the supported Sire unit names.
+    _sire_units = [
+        "kilogram",
+        "gram",
+        "milligram",
+        "microgram",
+        "nanogram",
+        "picogram",
+        "femtogram",
+    ]
+
+    # Dictionary of allowed units.
+    _supported_units = {
+        "KILOGRAM": _SireUnits.kilogram,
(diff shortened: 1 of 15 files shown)


python/BioSimSpace/Sandpit/Exscientia/Types/_mass.py
```diff
@@ -0,0 +1,403 @@
...
+        >>> import BioSimSpace as BSS
+        >>> length = BSS.Types.Length(12, "G")
+        >>> print(length.kilograms())
+
+        The same as above, except passing a string representation of the
+        length to the constructor.
+
+        >>> import BioSimSpace as BSS
+        >>> length = BSS.Types.Length("12 G")
+        >>> print(length.kilograms())
```
mb2055, line 132: Looks like the docstring here is wrong. Presumably it should be something like 
```python
mass = BSS.Types.Mass(12,"G")
print(mass.kilogram())
...
mass = BSS.Types.Length("12 G")
print(mass.kilogram())
````

Comment (mb2055) [reply python/BioSimSpace/Sandpit/Exscientia/Types/_mass.py:132]: Also, looks like the function `.kilogram()` not `.kilograms()`

File: python/BioSimSpace/Types/_mass.py [hunk 2]
```diff
@@ -0,0 +1,403 @@
...
+        >>> import BioSimSpace as BSS
+        >>> length = BSS.Types.Length(12, "G")
+        >>> print(length.kilograms())
+
+        The same as above, except passing a string representation of the
+        length to the constructor.
+
+        >>> import BioSimSpace as BSS
+        >>> length = BSS.Types.Length("12 G")
+        >>> print(length.kilograms())
```
mb2055, line 132: Same as above, docstring needs to be changed

mb2055 (changes requested):
Everything seems to work, just a few doc changes

python/BioSimSpace/Sandpit/Exscientia/Types/_mass.py
lohedges: Thanks, will fix this now. I think I copied the previous class template over, but clearly forgot to modify all of the docstring.

mb2055 (approved):
All good :+1:

PR #393, opened 2025-02-17
Fix issue #392

### LINKED ISSUE #392: [BUG] Use direct RDKit conversion when checking formal charge
Currently we go via PDB to RDKit when calculating formal charge. This isn't ideal, since it fails in many cases. We should switch to direct RDKit conversion now it's available.

lohedges:
This PR fixes #392 by improving the robustness of the internal formal charge calculation within the `Parameters` sub-module. Now we preferentially use the format the the original molecule was loaded from, either PDB or SDF, or try both if none are present in the file formats list. This means that we can still get a result if the PDB conversion doesn't work, etc.

Note that I haven't been provided with the problem file in this case, so haven't added an additional test. All existing tests pass.

### DIFF
--- a/python/BioSimSpace/Parameters/_utils.py
+++ b/python/BioSimSpace/Parameters/_utils.py
@@ -28,14 +28,14 @@ __all__ = ["formalCharge"]
 
 import tempfile as _tempfile
 
-from .. import _is_notebook
+from .. import _isVerbose
 from .. import IO as _IO
 from .. import _Utils
 from ..Units.Charge import electron_charge as _electron_charge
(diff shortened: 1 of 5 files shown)


lohedges:
Also note that I chose not to use direct conversion, since it appears to be less reliable if the molecule wasn't loaded from SDF, PDB, or Mol2, i.e. whatever RDKit does to fix things when going from those formats seems to more reliably recover the formal charge than converting directly when that info isn't present, e.g. if you went from a molecule loaded from AMBER files.

mb2055 (approved):
All good :+1:

PR #396, opened 2025-03-08
Fixed issues related to align_selection=None

CHENFENG1020:


CHENFENG1020:
BSS.Metadynamics.CollectiveVariable.RMSD raises an error when `align_selection=None`. However, the parameter information states that if None, the RMSD will be calculated without alignment.

## Suggested reviewers:
@lohedges @mb2055 

1. Would it be better if setting `align_selection` to None,  the system will be aligned using `rmsd_selection`?
2. I am working on ensuring  [AMMo](https://github.com/michellab/AMMo) is compatible with the latest BioSimSpace. @AdeleHardie uses this function to format the reference PDB for PLUMED in AMMo, and in her example, the reference contains fewer residues. I have read #376 and #379 and noticed that this function previously worked when the residue numbers differed between the reference and the system. Is it necessary to enforce that the system and reference have the same number of residues (and therefore the same number of atoms)? Could we instead ensure that `align_selection` and `rmsd_selection` each have the same number of residues and atoms in both the system and the reference?

lohedges:
Thanks, I'll take a look next week. The main issue is that the selections can now refer to atoms in multiple molecules, so the old logic for matching atoms from the reference to those in the system, which matched by residue index and atom name, is much more complicated. It's certainly possible, but I wasn't sure if it was worth the effort at the time. I'll see how easy it would be to implement.

Issue #397, opened 2025-03-20
[BUG] Nodes fail to run in interactive sessions when using custom output names


mb2055:
**Describe the bug**
Using the `file_prefix` argument in `node.validate` , then attempting to run the node within an interactive session with `BSS.Node.run` causes the following exeption: FileNotFoundError: [Errno 2] No such file or directory: 'output.yaml'.

This is due to the assumption[ here](https://github.com/OpenBioSim/biosimspace/blob/974cdd1d0c9d382764eb089b0ab0bfd3b02fca55/python/BioSimSpace/Node/_node.py#L131-L155) that the output file will always be called `output.yaml`, even if `file_prefix` is set to something different.

**To Reproduce**
1. Create a basic node e.g.:
```python
import BioSimSpace as BSS

node = BSS.Gateway.Node("A node.")

node.addInput("in", BSS.Gateway.Integer(help="in"))

node.addOutput("out", BSS.Gateway.Integer(help="out"))

node.setOutput("out", 1)

node.validate(file_prefix="test")
```
2. Attempt to run the node in an interactive session:
```python
In [1]: import BioSimSpace as BSS

INFO:numexpr.utils:Note: NumExpr detected 20 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 16.
INFO:numexpr.utils:NumExpr defaulting to 16 threads.

In [2]: BSS.Node.setNodeDirectory(".")

In [3]: BSS.Node.run("null_node", {"in":1})
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ in <module>:1                                                                                    │
│                                                                                                  │
│ /home/matt/code/biosimspace/python/BioSimSpace/Node/_node.py:160 in run                          │
│                                                                                                  │
│   157 │                                                                                          │
│   158 │   if proc.returncode == 0:                                                               │
│   159 │   │   # Read the output YAML file into a dictionary.                                     │
│ ❱ 160 │   │   with open("output.yaml", "r") as file:                                             │
│   161 │   │   │   output = _yaml.safe_load(file)                                                 │
│   162 │   │                                                                                      │
│   163 │   │   # Delete the redundant YAML files.                                                 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
FileNotFoundError: [Errno 2] No such file or directory: 'output.yaml'
```

**(please complete the following information):**
 - OS: Ubuntu 22.04.5 LTS
[...]

PR #398, opened 2025-03-21
Fix for issue #397

mb2055:
Fix for issue #397 .

This fix ensures that each node input file has a unique identifier, preventing conflicts when multiple node instances are running simultaneously. 

Fix for the output file is not ideal but I can't think of a better way to do it since we need to read the `file_prefix` from the `args` dictionary without knowing which of the arguments (if any) will be used to set the `file_prefix` when `validate` is called.  The solution here is simply to assume that this will be set using an argument called `file_prefix` and that if this argument does not exist then the default name of `output.yaml` is used. __It does not account for any situation in which the `file_prefix` is set using a value that is not the `file_prefix` argument__.

 Also adds a `getNodeDirectory` utility function that returns the current value of `_node_dir` .

### DIFF
--- a/python/BioSimSpace/Node/_node.py
+++ b/python/BioSimSpace/Node/_node.py
@@ -36,7 +36,7 @@ from sire.legacy import Base as _SireBase
 # Set the default node directory.
 _node_dir = _os.path.dirname(__file__) + "/_nodes"
 
-__all__ = ["list", "help", "run", "setNodeDirectory"]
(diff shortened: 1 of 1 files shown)


lohedges:
Sorry for the delay with this, have been trying to think of a cleaner solution. Two options that I can think of:

1) Allow the user to pass an ID or input file name to the `run` function.
2) Run the process within a temporary workspace, allowing the user to specify a `work_dir` option as a kwarg to `run`. This would necessitate the use of absolute paths for inputs, though, but this could be documented.

PR #400, opened 2025-05-27
Sync with Recursion remote

lohedges:
This PR synchronises with changes from the Recursion remote.

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/Process/_gromacs.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/Process/_gromacs.py
@@ -1930,6 +1930,29 @@ class Gromacs(_process.Process):
         """
         return self.getPressure(time_series, block=False)
 
+    def getDensity(self, time_series=False, block="AUTO"):
+        """
+        Get the Density.
+
+        Parameters
+        ----------
+
+        time_series : bool
+            Whether to return a list of time series records.
+
+        block : bool
+            Whether to block until the process has finished running.
+
+        Returns
+        -------
+
+        density : :class:`GeneralUnit <BioSimSpace.Units.Mass.kilogram / BioSimSpace.Units.Volume.meter3>`
+            The Density.
+        """
+        return self.getRecord(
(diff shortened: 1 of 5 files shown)


lohedges:
Having looked back at the CI, these errors also appeared in #398, so they aren't unique to this PR.

lohedges:
This has fixed the `ValueError` exceptions reported above, but I'm still seeing the following error on Linux Python 3.10 and macOS Python 3.11:

```python
   @pytest.mark.skipif(
        has_antechamber is False or has_tleap is False,
        reason="Requires AmberTools/antechamber and tLEaP to be installed.",
    )
    def test_acdoctor():
        """
        Test that parameterising negatively charged molecules works when acdoctor
        is disabled.
        """
    
        # Load the molecule.
        mol = BSS.IO.readMolecules(f"{url}/negative_charge.sdf")[0]
    
        # Make sure parameterisation fails when acdoctor is enabled.
>       with pytest.raises(BSS._Exceptions.ParameterisationError):
E       Failed: DID NOT RAISE <class 'BioSimSpace._Exceptions._exceptions.ParameterisationError'>
```

I assume that `antechamber` has been updated so that `acdoctor` is no longer required. It's annoying that this is platform dependent. It appears that _older_ Python variants have _newer_ versions on `ambertools`:

Linux Python 3.10 (fails):     
```
ambertools:                       24.8-cuda_None_nompi_py310h1ac786c_100 conda-forge         
```

Linux Python 3.11 (works):
```
    ambertools:                       23.6-cuda_None_nompi_py311h4a53416_105 conda-forge         
```

Not sure if I should just pin `ambertools` for now (other issues have been reported elsewhere for other dependencies) or just skip the test for now. Annoyingly, as the other error has shown there's no way to determine the version of `antechamber` so we can't conditionally change the xfail based on the version.
```

**Comment by lohedges** [comment]
`sander` does report the version information, so maybe we could use that as a proxy, i.e.:

```
sander -v
sander: Version 24.0
```

lohedges:
Bah, that works locally in a conda environment with `ambertools`23 and 24, but doesn't work in the CI since `sander` doesn't report any version info either. I'll just skip the test for now.

Issue #426, opened 2025-07-22
[BUG] Certain ParmEd versions are incompatible with recent NumPy versions


lohedges:
ParmEd has not correctly pinned NumPy so certain versions are incompatible, which triggers the following exception:

```python
ModuleNotFoundError: No module named 'numpy.compat'
```

See [here](https://github.com/ParmEd/ParmEd/issues/1406) for details. I'm not sure if/when this will be resolved, so we will probably need to do an in-place import of `parmed` to avoid breaking _all_ BioSimSpace functionality. (It is only used in the AMBER FEP squash code and as an OpenMM setup workaround.) Annoyingly this would mean that the exception might only be triggered late in a workflow, therefore wasting compute.

lohedges:
Another workaround for this is to pin to an old version of `numpy`, e.g. 1.26.4. The BioSimSpace environment is resolvable with this, but it could cause issues going forward. I'll consider adding this if users are being tripped up by the above exception triggering during their workflows.

PR #428, opened 2025-07-22
Fix issues #426 and #427

### LINKED ISSUE #427: [BUG] Don't explicitly require AnteChamber when OpenFF-NAGL is installed
When OpenFF-NAGL is present, there is no need to check for AnteChamber, since it isn't required.

lohedges:
This PR fixes closes issue #426 and closes #427. The fix for #426 isn't ideal, but is probably all we can do until the issue is properly fixed upstream.


xiki-tempula (approved):
The change looks sensible but is not my issue

PR #440, opened 2025-09-08
Add support for renumbering molecules when copying a system

### LINKED ISSUE #439: System.copy() to renumber all molecules within the system
Currently the `Molecule.copy()` method can be used to obtain a copy of a molecule with a different (unique) `MolNum`. It would be good if `System.copy()` could be used to renumber all molecules within a system.

### LINKED ISSUE #443: Open Force Field parameterisation currently broken when openff-nagl isn't installed
Due to a bug in the latest `openff-interchange` package, our OpenFF parameterisation pipeline no longer works when the `openff-nagl` package isn't installed. This package is an optional dependency for us, i.e. NAGL charges will be used when installed, otherwise AM1BCC charges will be computed using `antechamber`.

The bug has been reported [here](https://github.com/openforcefield/openff-interchange/issues/1309) and fixed [here](https://github.com/openforcefield/openff-interchange/pull/1319). An updated `openff-interchange` package will likely be available in a week or so. For now, simply install `openff-nagl` alongside BioSimSpace when creating an environment, e.g.:

```
create -n openbiosim -c openbiosim/label/devel -c conda-forge biosimspace openff-nagl
```

lohedges:
This PR closes #439 by adding a `renumber` kwarg to the `System.copy()` method. When `True`, molecules in the system are given a unique number. The default behaviour is `renumber=False`, for consistency with the existing code and for performance.

Closes #443

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
@@ -260,6 +260,47 @@ class System(_SireWrapper):
         """Return the number of molecules in the system."""
         return self.nMolecules()
 
+    def copy(self, renumber=False):
+        """
+        Return a copy of this System.
+
+        Parameters
+        ----------
+
+        renumber : bool
+            Whether to give the copied molecules unique molecule numbers.
+
+        Returns
+        -------
+
+        System : :class:`System <BioSimSpace._SireWrappers.System>`
+            A copy of the object.
+        """
+
+        if not isinstance(renumber, bool):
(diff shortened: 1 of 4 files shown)


lohedges:
That didn't work since some ABFE tests for SOMD in the sandpit use hardcoded charge values in the pert files checks. I've adjusted these and hope that the NAGL charge delta is consistent across platforms.

lohedges:
This problem has actually been reported [here](https://github.com/openforcefield/openff-interchange/issues/1309), so I'll hold off on merging until we know whether this is an expected change from the `openff` side.

PR #456, opened 2025-10-07
Fix issue #445

### LINKED ISSUE #445: [BUG] Trajectory.getFrame/s fails when a perturbable system is used as the reference
Sometimes it appears that `Trajectory.getFrame/s` will fail if an AMBER topology file is used as an intermediate to reconstruct a frame from a GROMACS trajectory file. This can happen when the user passes a reference system for the topology, since the format associated with the _system_ will be used for the intermediate, rather than that of the trajectory. For example, this would happen if the system was created via AMBER force field parametrisation, but a simulation was then run with GROMACS.

Although it's not clear why this happens, an obvious solution would be to try reconstructing the system using the format of _both_ the reference system and simulation engine, e.g. within a `try/except`, using one, then the other as a fallback. An exception would only be raised when both fail.

lohedges:
This PR closes #445 by correctly handling perturbable reference systems. There are multiple subtleties:

* The reference might have molecular properties that mismatch the frame, e.g. `coordinates0` rather than `coordinates`.
* The number of atoms in the trajectory might not agree with the reference system, i.e. if the trajectory was from and AMBER simulation, in which case the trajectory contains a "squashed" representation. In this case the reference needs to first be squashed to match the trajectory, then unsquashed once the specific frames of interest have been loaded.

I have tested this locally using RBFE output for AMBER, GROMACS, and SOMD, using all trajectory backends. I'm not sure what to do about unit tests, since I don't currently have a consistent test case for all engines. (AMBER isn't working locally.) Ideally I'd have a small 10 frame vacuum system simulated with all engines. I'll see if I can generate this, then update accordingly.


lohedges:
I've managed to get the non CUDA PMEMD working, so will generate some test files and update.

PR #466, opened 2025-10-27
Switch to new Sire API and refactor imports

lohedges:
This PR switches introduces two changes:

1) Switching fully over to the new Sire API. This means that Sire ca now be imported _before_ BioSimSpace, allowing BioSimSpace functionality to be used within Sire itself. To avoid immediate activation of the new API, I've updated BioSimSpace to use lazy importing of sub-modules using the same approach as Sire.
2) The import structure of BioSimSpace has been restructured to move from module level imports (generally at the top of a module) to in-place use, i.e. imports are performed in functions and methods where they are used. Along with 1), this helps reduce the import overhead of the code. Note that the change could lead to import errors not being caught until late in a workflow, but we should hopefully catch these with our CI, e.g. changes to external API calls, etc.

With the changes above, the average time for importing BioSimSpace (over 10 trials) is around 90 ms on my laptop. Prior to this it was around 3100. (Importing into a sub-module still has a performance hit, but the cost is now cumulative, rather than paid once up-front.) Hopefully this will also make our test-suite quicker to run.

I've tested locally but some of the tests don't run due to `antechamber` being broken in my environment. I'm expecting that I'll need to fix minor issues following the CI runs.

### DIFF
--- a/python/BioSimSpace/Align/_align.py
+++ b/python/BioSimSpace/Align/_align.py
@@ -2440,6 +2440,9 @@ def _draw_molecules(
         molA_atom_idx, molB_atom_idx]
         default None
     """
+    # Adapted from GUFE: https://github.com/OpenFreeEnergy/gufe
+    # Licensed under the MIT license.
+
     from rdkit.Chem import Draw
     from rdkit.Chem import AllChem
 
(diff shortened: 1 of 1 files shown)


mb2055 (approved):
I've run through all of the website tutorials, plus the BSS_tutorial suite and some of my own scripts and everything seems to be working well :+1:

PR #502, opened 2026-02-24
Fix ring break detection

lohedges:
This PR fixes issues in the detection of ring-breaking and/or size change transformations during molecule merging.


lohedges:
This has already been reported [here](https://github.com/conda-forge/sympy-feedstock/issues/67) with a fix incoming [here](https://github.com/conda-forge/sympy-feedstock/pull/68).

PR #522, opened 2026-06-25
Feature custom roi maps

akalpokas:
This PR adds `custom_roi_map` functionality to `matchAtoms()` function as well as the underlying `_roiMatch()` function in order to enable user defined mappings for ROI regions. This allows for creation of protein perturbations involving advanced use-cases, such as ring-breaks for example.  The argument only works when `roi` is passed to `matchAtoms()`, so the underlying behaviour of the function outside of protein FEP remains unchanged.

For example as detailed in the tutorial,  `custom_roi_map={}`  can be used to return an empty mapping for ROI region:
```python
mapping = BSS.Align.matchAtoms(molecule0=wt, molecule1=mut, roi=[15], custom_roi_map={})
BSS.Align.viewMapping(wt, mut, mapping, roi=15, pixels=500)
```
image

Or pass a custom mapping where the RDKit MCS algorithm would fail otherwise:
```python
mapping = BSS.Align.matchAtoms(molecule0=wt, molecule1=mut, roi=[15], custom_roi_map={204:204,205:205,203:203,202:202,211:208,206:206,213:210,207:207,214:211,210:213})

BSS.Align.viewMapping(wt, mut, mapping, roi=15, pixels=500)
```
image

___

Specifically this PR:
1. Adds the `custom_roi_map` functionality as described above.
2. Introduces `test_empty_custom_roi_mapping` and `test_custom_roi_ring_break_merge` test cases in [test_align.py](https://github.com/akalpokas/biosimspace/blob/feature_custom_ROI_maps/tests/Align/test_align.py) that test the following:
- `test_empty_custom_roi_mapping`: that a user can pass an empty `custom_roi_map` dictionary to return two unmapped ROI residues, while the rest of the protein is mapped correctly. This allows for setting up transformations using a "dual-topology" approach.
-  `test_custom_roi_ring_break_merge`: that a user can pass a custom mapping to create an alchemical ring transformation in a test system. The tests checks for this by asserting the number of alchemically created bonds as part of the mapping is equal to 1 (Leucine to Proline transformation is reverse of a ring-break).
3. Updates the [protein FEP tutorial](https://github.com/akalpokas/biosimspace/blob/feature_custom_ROI_maps/doc/source/tutorials/protein_mutations.rst) to demonstrate the new functionality

* I confirm that I have added a test for any new functionality in this pull request: y
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: y

## Suggested reviewers:
@lohedges

This PR is non release blocking and can be merged after the release if needed.


lohedges:
Thanks for this. Looks like there are some weird failures due to some OpenFF packages. Will investigate tomorrow.

src/BioSimSpace/Align/_align.py
```diff
@@ -1413,22 +1438,29 @@ def _roiMatch(
...
 
-        # Look up the absolute atom indices in the molecule
-        res0_lookup_table = list(mapping.keys())
-        absolute_mapped_atoms_res0 = [res0_idx[i] for i in res0_lookup_table]
+        # Look up the absolute atom indices in the molecule if not using a custom ROI mapping.
+        if custom_roi_map is not None:
+            absolute_roi_mapping = custom_roi_map
+        else:
+            res0_lookup_table = list(mapping.keys())
+            absolute_mapped_atoms_res0 = [res0_idx[i] for i in res0_lookup_table]
```
lohedges, line 1468: Can we validate the `custom_roi_map` here. Firstly, if it's not `None`, then we need to check that it's a dict, and raise a `TypeError` if not. If it is a dict, then we also need to validate that the key:value pairs are in bounds, i.e. the key is in the range 0 to `molecule0.nAtoms()` and the value is in the range 0 to `molecule1.nAtoms()`.

src/BioSimSpace/Align/_align.py
```diff
@@ -875,6 +887,7 @@ def matchAtoms(
             molecule0=molecule0,
             molecule1=molecule1,
             roi=roi,
+            custom_roi_map=custom_roi_map,
```
lohedges: Can these be passed as positional arguments, since that is how they are defined in the function definition below, i.e. everything up to `prun_perturbed_constraints` is an arg.

tests/Align/test_align.py
```diff
@@ -671,6 +671,79 @@ def test_roi_flex_align(protein_inputs):
...
+    roi_res_idx = [a.index() for a in wt.getResidues()[15].getAtoms()]
+    mapping = BSS.Align.matchAtoms(
+        molecule0=wt, molecule1=mut, roi=[15], custom_roi_map={}
+    )
+
+    # check that the mapping does not contain any atoms of the region of interest of WT protein
+    for atom_idx in roi_res_idx:
+        assert atom_idx not in mapping.keys()
+
+@pytest.mark.skipif(has_amber is False, reason="Requires AMBER and to be installed.")
```
lohedges: Just a small typo. Should be `Requires AMBER to be installed."

lohedges (changes requested):
Thanks for this, the tutorial is nice and clear. Just a few minor changes.

src/BioSimSpace/Align/_align.py
akalpokas: I've added the `custom_roi_map` validation now. Just to note that to validate the custom mapping, we need to check that keys and values are within `res0_idx` and `res1_idx` respectively (absolute residue atom indices as if they were within the full protein, rather that range 0 to `molecule0.nAtoms()`. This is because the `custom_roi_mapping` is not about replacing the MCS calculated mapping for two standalone residues, but rather their full mapping as if they were in a protein. Otherwise if the user wanted to use something like PyMol/VMD to map these residues manually, they would need to extract the residues from the proteins themselves.

I added an extra test that checks whether the user provided mapping is valid by checking for above.

lohedges: Ah yes, of course, an atoms bound check isn't fine-grained enough, this makes sense.

lohedges (approved):
Many thanks!

